PR #14, opened 2020-03-16
Issue10

diegogentilepassaro:
@santiagohermo here you have the #10 for peer review. Note that on top of the steps described in the issue I've also fixed the R library. :D

Let me know your thoughts!


analysis/output/make.log
```diff
@@ -95,866 +87,1278 @@ Notes:
...
             Absorbed FE | Categories  - Redundant  = Num. Coefs |
 ------------------------+---------------------------------------|
-                zipcode |      2379        2379           0    *|
-   calendar_month#state |       576           0         576     |
-             year_month |       112          12         100     |
-       year_month#state |      3085         576        2509    ?|
+                zipcode |      2381        2381           0    *|
+                    msa |       223           0         223     |
+   calendar_month#state |       594          31         563     |
+             year_month |       117          12         105    ?|
```
santiagohermo: Good to cluster at the `msa` level! This also suggest we should not cluster at the zipcode level, right?

analysis/code/preclean_event_studies.do
```diff
@@ -0,0 +1,86 @@
...
+end
+
+program prepare_data
+    syntax, time_var(str) geo_unit(str)
+
+    gen date = dofm(year_month)
+	gen calendar_month = month(date)
+	drop date
+	
+	replace min_event = 0 if dmin_actual_mw < 0.5
```
santiagohermo: Just flagging that in a future task we should add % variables, and of course try different percentages. For now looks good!

lib/R/library.R
```diff
@@ -1,122 +1,5 @@
...
-    writeLines("=========================================================================================\n", logfile)
-  }  
-  
-  print("Log file generated successfully.")
-  close(logfile)
-}
-## ================================= ##
-
+this.dir <- dirname(parent.frame(2)$ofile)
+setwd(this.dir)
```
santiagohermo: Nice trick papito! A world of care: we should try running previous R stages and make sure this `setwd()` does not mess up the running there.

analysis/code/event_study.do
```diff
@@ -2,90 +2,25 @@ clear all
...
-	use `instub'zipcode_yearmonth_panel.dta, clear
+	foreach window in 12 24 {
+		create_event_plot, depvar(rent2br_median) event_var(rel_months_min_event`window')      ///
+			controls(" ") window(`window') ///
+			absorb(zipcode msa calendar_month##state year_month)
 
-	prepare_data, time_var(year_month) geo_unit(zipcode)
+		create_event_plot, depvar(rent2br_psqft_median) event_var(rel_months_min_event`window')      ///
+			controls(" ") window(`window') ///
+			absorb(zipcode msa calendar_month##state year_month)
```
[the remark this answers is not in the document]
gabrieleborg: In the light of what discussed today, we should re-assess what to use, but the correct one here is maximum as you say !

analysis/code/preclean_event_studies.do
```diff
@@ -0,0 +1,86 @@
...
+	forvalues i = 2(1)`window_span' {  					
+		bysort `geo_unit' (`time_var'): replace rel_months_`event_dummy' = `i' ///
+		    if event_start_non_overlap[_n - `i' + 1] == 1
+	}
+	replace rel_months_`event_dummy' = 1000 if rel_months_`event_dummy' == .
+
+
+	drop event_start event_start_non_overlap
+	rename rel_months_`event_dummy' rel_months_`event_dummy'`window'	
+end
```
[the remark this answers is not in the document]
diegogentilepassaro: This is a great idea! Let's do it next time that we create event study plots.

analysis/output/make.log
diegogentilepassaro: I actually clustered at the zipcode and added msa fixed effects. In fact, your comment made me spot a mistake: I was clustering at the zipcode in a way that was hardcoded into the library. I will go ahead and fix that now and commit again a fresh run. 

I believe that in a follow up task we should create a figure that is a panel of the same event-study (say the one for the latest MW change) with progressively more controls. Like first just two-way FE (zipcode and month), to the left add state##year, and so on to understand whihc is the source of variation that matters the most and think about what else could we control for. Maybe when we do what Jesse said and run the events in first differences things change a bit.

analysis/code/preclean_event_studies.do
diegogentilepassaro: Yes! And we should also directly run a regression model with the first diff of the MW level on the LHS.

diegogentilepassaro:
@santiagohermo I am not sure of what is going on with the plot that you point out [here](https://github.com/diegogentilepassaro/min_wage_rent/pull/14#pullrequestreview-375608394). :/

I hope that we can figure that out soon in a future task!

analysis/code/preclean_event_studies.do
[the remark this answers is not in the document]
santiagohermo: Sounds good! I added a [note](https://github.com/diegogentilepassaro/min_wage_rent/wiki/Outline#implementation) to the wiki.

analysis/output/make.log
santiagohermo: Thanks! I still don't understand that small box down there haha. You mean your are adding a cluster option [here](https://github.com/diegogentilepassaro/min_wage_rent/blob/24d5ad7e3695a570e9f504be4feb9ae4835c4a15/analysis/output/make.log#L919-L923) to choose the cluster variable when running the plot? Makes sense to me!

I think that running the plot with progressively more controls is a great idea. We should definitely try that out in the future.

diegogentilepassaro: Yes! I mean that in the lib function/create_event_plot.ado I will add an option to declare the level of clustering  :D

lib/R/library.R
diegogentilepassaro: Actually this was messing things up! And it might be the reason why base was breaking (@gabrieleborg FYI). I will actually go for just leaving /lib/R/functions and inside that folder each functions will live and each script that needs it will need to declare it specifically.  This is the spirit of R when you have to declare a library. An alternative would be to build an R library. We can do this in the future.

analysis/output/make.log
santiagohermo: Awesome!

lib/R/library.R
santiagohermo: It is a complicated problem. Thanks! I'll mark this as resolved

analysis/code/preclean_event_studies.do
```diff
@@ -0,0 +1,86 @@
...
+	bysort `geo_unit': egen last_`event_dummy'_`time_var' = max(`event_dummy'_`time_var')
+	gen last_`event_dummy'_rel_months`window' = `time_var' - last_`event_dummy'_`time_var'
+	replace last_`event_dummy'_rel_months`window' = last_`event_dummy'_rel_months`window' + `window' + 1
+	replace last_`event_dummy'_rel_months`window' = 0 ///
+	    if last_`event_dummy'_rel_months`window' <= 0
+	replace last_`event_dummy'_rel_months`window' = 1000 ///
+	    if last_`event_dummy'_rel_months`window' > `window_span'
+		
+	drop `event_dummy'_`time_var' last_`event_dummy'_`time_var'
+end
```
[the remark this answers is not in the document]
diegogentilepassaro: Great idea @gabrieleborg! We should do that on the next iteration.

Issue #18, opened 2020-03-21
Some updates to the repo


santiagohermo:
In line with our old meetings, in this repo I am going to make some changes to the stucture of the pipeline. 

In particular, I will
1. Change online folder from "raw_data" to [this](https://drive.google.com/drive/u/1/folders/1PRXhH-6Ny_jNGUcS7vjb0A-QzfNEt816) broader folder,
2. within the new Drive folder, I will implement derived_large to save large files we don't to run over and over again,
3. I will add separate output folders -"descriptive" and "event-studies"- to make room for different kind of analyses.

santiagohermo:
Thread follows in PR #19

santiagohermo:
Changes merged to master in https://github.com/diegogentilepassaro/min_wage_rent/commit/5290bc35e5f4c32a915118afb01506bf98a959af

As part of the review process we learnt that `save_data` breaks with data type `integer64` (columns with integer values larger than `2147483647`).

Issue #13, opened 2020-03-09
Uncover Zillow Index Methodology


gabrieleborg:
**Goal**
Have a clear understanding on how the Zillow variables are built, what they measure, how they can be used and what's the value for our study

**Steps**

1. Read and download Zillow methodology files
2. Figure out the best potential variables for our designs, and why. 
3. Report exactly what the variables are, and  how they are computed.
4. Think and report potential consequences of the variable build on our computations.

**Notes**
1. Smoothed (moving averages) variables should have a progressive dynamics when impacts are discrete. Jumps on impact shouldn't be possible. 
2. Maybe a follow-up task, once we understand what we are going to use, would be to perform Montecarlo simulations to understand the impact of smoothing on a toy series that resembles the data.
3. Think about seasonally adjustments in the y variable.

santiagohermo:
There is a lot of information in [that file](https://github.com/diegogentilepassaro/min_wage_rent/issues/13#issuecomment-599603633) @gabrieleborg. So much so that I think we should include it either in the repo or google drive folder. What do you think @diegogentilepassaro ?

Zillow is doing a lot of things to the raw data to obtain those indices. I think our main results should not come from them, the moving average is kind of a killer! Rather, we should use Zillow's raw data (and other sources). For example, the `InventoryMeasurePublic`, `MedianListingPrice` and `MedianRentalPrice` look promising. Of course, we should try the indices anyway to make sure the results are as expected (they should look like the main results but attenuated, right?).

Question: do we have access to the data on "market segments" that you mention in the file?

Regarding the formatting:
- I can't open links from the file. I think it's because overleaf modifies them (e.g., https://www.overleaf.com/project/5e6a9dfb1a5a40000198775a#https://www.zillow.com/info/whats-the-zillow-home-value-index/)
- Sometimes there are question marks. Is that you not being sure of what Zillow did or some broken reference?

Thanks again!

gabrieleborg:
Hi @santiagohermo sorry for the delay. I agree in that we should eventually use raw measures such as median prices or rents. I think we could keep the indexes as robustness check to validate some assumptions (e.g. assumption regarding seasonality or those things we know are addressed in the index construction). 

Yes, we do have data on different market segments. These are simply the combination between property type (as classified in the pdf as well) and different number of bedrooms. 

Question marks are my personal doubts. I left them when I couldn't properly understand what was going on!

gabrieleborg:
[...]

PR #19, opened 2020-03-22
PR for #18: Some updates to the repo

santiagohermo:
Can you take a quick look at these changes @gabrieleborg @diegogentilepassaro ? 

Updates:
- Now the repo includes a `/drive/` folder linked to [this GDrive](https://drive.google.com/drive/u/1/folders/1PRXhH-6Ny_jNGUcS7vjb0A-QzfNEt816). `/raw_data/` is a subfolder of drive.
- I updated `save_data` (I discussed it in on of Jesse's projects with another RA). data manifests now look nicer: [example](https://github.com/diegogentilepassaro/min_wage_rent/blob/52d3d14133cde328f178c198bd92efec7c982214/derived/output/data_file_manifest.log)
- `base` still outputs to `base/output/`. However, it takes files from `/drive/raw_data/`
- `derived` takes files from base and outputs into `/drive/derived_large/output/` ([link](https://drive.google.com/drive/u/1/folders/1b1mm3RvbYxA5Pir3ik2hFr6ZWU1dk94K) to files).  Data file manifests still live in `derived/output/`. We now also compute min wage changes as a percent (see [here](https://github.com/diegogentilepassaro/min_wage_rent/blob/52d3d14133cde328f178c198bd92efec7c982214/derived/output/data_file_manifest.log#L285-L287))
- `analysis` consists of two folders:  `analysis/output/event_study` and `analysis/output/descriptive`. The second one is intended to be used in #13.

Finally, let me insist in using a local copy of the GDrive to make changes in the repo. (e.g., I had to go through the newly added variables in #13 and change names in analysis to make it work. This should be added to the drive after #13 is merged.)

Let me know if you have any questions or find anything weird.


.gitignore
```diff
@@ -1,12 +1,14 @@
-raw_data/
-raw_data
```
diegogentilepassaro: This line should be kept! In mac and Windows .gitignore files work slightly different and without the / the ignore works in Mac.

.gitignore
```diff
@@ -1,12 +1,14 @@
...
-analysis/temp/*
-base/temp/*
-derived/temp/*
+/raw_data/
+/drive/
+/analysis/temp/
```
diegogentilepassaro: As we split analysis/output in two we should also split the code in two. This is mainly because of computational reasons. One way would be to do two folders within analysis (event_study and descriptive) and in each case we have code, output and temp inside. If we go this way we need to be 100% to not introduce circularities in the code. For that a good safety procedure would be for analysis/some_folder to have inputs that only come from derived. If we go this way then this line should be changed to two lines: 

- analysis/event_study/temp

- analysis/descriptive/temp

derived/code/ReshapeMergeZillow_zip.R
```diff
@@ -1,6 +1,6 @@
-# Preliminaries
-source("../../lib/R/functions/save_data.R")
-source("../../lib/R/functions/load_packages.R")
+remove(list = ls())
+source("../../lib/R/library.R")
```
diegogentilepassaro: Why we changed the library again? I didn't understood that we were changing it back again after our discussion per #14. Is this because we didn't updated from master properly? Or is it intentional?

derived/code/ReshapeMergeZillow_zip.R
```diff
@@ -18,17 +18,23 @@ reshape_zillow_zip_level <- function(infiles){
    filenames <- filenames[str_detect(filenames, "Zip_*")]
    filenames <- filenames[!str_detect(filenames, "_Summary.csv")]
    
-   value_names <- c('rent2BR_median', 'rent2BR_psqft_median', 'rent_psqft_median_SFR', 'zhvi2BR')
+   value_names <- str_replace_all(filenames, "Zip_", "")
+   value_names <- target_varname(value_names)
```
diegogentilepassaro: Why do we need this line?

lib/R/library.R
```diff
@@ -0,0 +1,136 @@
+## =========== load_packages =========== ##
```
diegogentilepassaro: This way of building the library is not great. First, it is pretty cluttered so changing/adding/removing a single function is harder than it should be. Second, it involves loading a ton of things each time we invoke it, when instead we probably need loading one or two. Now this things don't make much difference but in the long run they might.

.gitignore
gabrieleborg: I think this should be the way. as Long as tasks are clearly separate (such as descriptive and event studies for example) I believe we will gain in procedural clarity in applying a task-specific structure in `analysis/`.

derived/code/ReshapeMergeZillow_zip.R
gabrieleborg: which one are you referring to specifically? Line 22 implements  a function that takes care of cleaning and formatting zillow series' names. This must be added to line 21 if we are actually selecting by default all csv files in a folder (e.g. if I want to build the descriptive table for _all_ Zillow series I don't want to manually change every name).  As you can see in the deleted line 21, once we settle on few target series then it'll make sense to select them manually and rename them as we prefer.

derived/code/ReshapeMergeZillow_zip.R
gabrieleborg: So far the two main _higher-level_ R functions we need are these two. The latter to load script-specific packages and the former to save data. I think it is ok to load them together within the `library.R` script. I am not following entirely this changing neither ahah, but I believe this is a robust setting.

santiagohermo: I apologize for changing it. I got carried away by the upgrades in `save_data` and forgot about our discussion. Per your other comment [here](https://github.com/diegogentilepassaro/min_wage_rent/pull/19/files?file-filters%5B%5D=.R&file-filters%5B%5D=.ado&file-filters%5B%5D=.do&file-filters%5B%5D=.log&file-filters%5B%5D=.md&file-filters%5B%5D=.png&file-filters%5B%5D=.py#r396113760), I changed it back to calling each function separately. Let me know if that's better.

.gitignore
santiagohermo: I restructured following diego's comment @gabrieleborg. I believe the way he suggested is more inline with the overall structure of the repo, and it makes it very simple to add new analysis in the future. Now structure is as follows:
> analysis
> --- ./event_study
> ------- ./code (with make.py)
> ------- ./output
> ------- ./temp
> --- ./descriptive
> ------- ./code (with make.py)
> ------- ./output
> ------- ./temp

 Let me know what you think!

diegogentilepassaro:
I am reviewiing this right now! :D

Issue #7, opened 2020-03-03
Add Recent Minimum Wage Data


gabrieleborg:
**Goals**
Add both state and local (county and city) level minimum wage data for the period Aug 2016 - Dec 2019. 

**Sources**
- State-level changes: [link](https://www.ncsl.org/research/labor-and-employment/state-minimum-wage-chart.aspx)
- Local-level changes: [link](http://laborcenter.berkeley.edu/minimum-wage-living-wage-resources/inventory-of-us-city-and-county-minimum-wage-ordinances/)

**Steps**
1. Add state changes to a state-level xslx file with the following columns (leave empty if not applicable): statefips, statename, year, month, day, VZ_mw, VZ_mw_healthinsurance, VZ_mw_smallbusiness
2. Add local changes to a local-level xlsx file with the following columns (leave empty if not applicable): statefips, statename, locality, year, month, day, VZ_mw, VZ_mw_tipped, VZ_mw_healthinsurance, VZ_mw_smallbusiness, VZ_mw_smallbusiness_mincompensation, VZ_mw_hotel
3. Adjust the dofiles in base to build the state and substate panel of minimum wage changes

santiagohermo:
Per call today I will collaborate with this task. After this and #13 are done we will move on to new tasks.

santiagohermo:
Interestingly, MW changes by counties in Iowa are not enforceable since March 30, 2017. This affects, for example, Johnson County: https://www.johnson-county.com/dept_supervisors.aspx?id=18549
The county continues to recommend changes in the MW. It may be interesting to see if those also have an effect. I did not add them to our data. FYI @diegogentilepassaro @gabrieleborg

santiagohermo:
It looks like we are missing some zipcodes in Oregon, where there are three MW: Portland, Portland Metro Area, Oregon State. We have no data for the Portland Metro Area. ([link 1](https://squareup.com/us/en/townsquare/your-guide-to-oregon-minimum-wage), [link 2](https://www.oregonlive.com/business/2017/07/oregon_minimum_wage_hike.html))

santiagohermo:
I am done with the cities/counties. The number of MW increases quite a bit, from 171 to 308. What I did is the following:
[...]

PR #20, opened 2020-03-23
PR for #7: Add recent MW data

diegogentilepassaro:
Time to PR this issue! 

@santiagohermo probably it is a good idea that you review my work and I review yours. After we closed #19 I have updated the issue #7 branch with master and I have committed a fresh run. There I have implemented some minor changes to base and derived that you should please review as part of this PR. The only not obvious change that I've implemented was to remove a part under \derived\label_final_dta in which we were eliminating zipcode-months without min-wage information. This was causing the panel to be unbalanced and we need a balanced panel for our event-studies to properly estimate FEs. Also, this shouldn't be happening as all zipcodes are assigned to some county which belongs to some state that either has a MW of its own or it has the federal one. This made me think that we need to think deeper about (i) the zipcode-county match (for example, there are also zipcode-months that we don't assign to a city because city and placename don't coincide). Probably, a good follow-up task although might be possible to improve during this PR.

Also, we should think deeper of (ii) what is the MW event that we consider. Is it the min, mean or the max? Do we include the special MW cases? What to do with the negative MW changes (when the binding MW goes down)? Another potential follow-up task that we can start thinking about how to make better in this PR. This matters in the code in this branch because we are using only min_events of at least $0.5. A question about this last thing: did we changed the analysis as per the changes in derived per [this comment](https://github.com/diegogentilepassaro/min_wage_rent/issues/7#issuecomment-601297259)?

The main PR work here is to check each others work:

- Do the changes in zipcode-months are consistent with the changes that we did? 

- Do we have outliers? Did we added them now or they where here before? (these together with (i) and (ii) might help understand the bumps)

- Did we miss some county or state changes? 

- Do the zipcodes that we are dropping  for the event-studies (either because they are missing or because they are not in an MSA) make sense? 

Feel free to add other questions that are important and I might be missing. Also, @gabrieleborg feel free to check our work and give your thoughts on the questions above!


gabrieleborg:
I haven't run it yet cause I'm having trouble with my laptop today. My good to go on the review pertains simply to code. I will check Diego's PR tasks as soon as I fix the issues I'm having.

santiagohermo:
I reviewed the state changes @diegogentilepassaro. All the ones I reviewed were correct. I also:
- added some of the MW changes after 31 dec 2019, since we may want to use them later.
- highlighted cells that add new information to the original version of the file (as in the substate level file).

I noted that many states have a tipped minimum wage which we are not adding (for example, [colorado](https://www.colorado.gov/pacific/cdle/minimumwage)). Should we add it? (fyi @gabrieleborg)

---
A couple of extra comments:
- Another interesting source for MW changes is https://www.laborlawcenter.com/state-minimum-wage-rates/
- Per our discussion in the last call we may want to add a dummy indicator that equals 1 if the MW change was pre-announced. We can tackle this in a follow-up issue in the future, for now I added it to [the outline](https://github.com/diegogentilepassaro/min_wage_rent/wiki/Outline#data)

gabrieleborg:
Thanks @santiagohermo for your comments. I don't think we need to add tipped minimum wage, and [here](https://github.com/diegogentilepassaro/min_wage_rent/pull/23#issue-395367837) I explain why. 

I started thinking how to deal with pre-announced changes as well. It will involve some reasoning though so I agree on putting it in the outline for now and open a new issue in the future

gabrieleborg:
Let's merge this? @diegogentilepassaro @santiagohermo

diegogentilepassaro:
Awesome work @santiagohermo! Good idea to add to the outline the tipped wage. 

Just finished reviewing some instances of your substate additions and they also look good. I will go ahead and merge this.

PR #22, opened 2020-03-25
PR for #13 - uncover zillow methodology

gabrieleborg:
@santiagohermo @diegogentilepassaro I revised a bit the code and, as discussed, I am moving this task to PR. 

### General Comments
Both the descriptive table and the plots are created with a single R script living in `analysis/descriptive/code`. Outputs are located in `analysis/descriptive/output`.

First, I already updated many of the Zillow series up to Dec 2019. In this way the descriptive table returns an informative picture of the variables we are actually going to use. 

image

The blue shadowed variables relates to house listing prices and home value, while the orange-shadowed ones to rents. Within each color group, the light colored ones refer to raw series of median values at zip-code level, while the dark colored ones to Zillow indexes. I have highlighted those series that I believe are the best suited for our goals: The main one I would use is **medrentprice_SFCC**, which is the median rent price for Single-family houses and Condominium and Cooperative houses. As you can see it's one of the most populated series for rent (tot_zip).  Here is it's average value time-series from the plot function: 

image

There are few observations in the initial years, but from 2011 it seems pretty stable in terms of size and its shape actually matches the house value curves (in the comparison tasks it will be important to compare it with the Fair Market Rent Index). Another series that we may want to keep into consideration is **medrentprice_Mfr5plus**, which refers to multi-family houses with 5 or more units (apartments). It is the most populated rent series, hence suggesting many buildings of this type are up for rent on Zillow. 

image

Its plot shows some issues especially in the first year of data, but beyond that it seems the variables gets fairly populated. Lastly, just as a reference I have highlighted the **Zillow Rent  Index (zri_SFCCMF)** for all possible type of house. This is one of the most comprehensive index (SF = single family house, CC = condos/coop, MF = multi-family house) so I think it'd be good to keep it under observation mostly as a variable to check assumptions on seasonal adjustment and smoothing, since we now know exactly how this is performed on such series. 

image

### Issues

For some reason I haven't figured out yet, the script returns a copy of the plots file also in the code folder, naming it **Rplots.pdf**.


diegogentilepassaro:
As per meeting: 

1. Keep all variables. Probably, we chose 4 or 5 for baseline panel, and the rest we save in bigger panel that has all variables. 

2. For the baseline panel, choose a fixed set of zipcodes or start analysis in a given year. For example, for the 2bdr rent variable the number of zipcodes per year evolves like:

Screen Shot 2020-04-03 at 4 59 18 PM

diegogentilepassaro:
@gabrieleborg, since now we have open #24 for creating a balanced panel for a subset of zipcodes (our baseline panel) that contains only the main outcome variables, the only outstanding things on this task seem to be:

1. Look into why the x-axis of some plots are wrong. 

2. Solve conflicts to merge to master. 

@santiagohermo let me know if I am forgetting something, but once these are addressed we should be able to merge and close this issue right?

santiagohermo:
Sorry for missing [this comment](https://github.com/diegogentilepassaro/min_wage_rent/pull/22#issuecomment-609482127) @diegogentilepassaro ! I will add to your list that we still have to add the complete table with summary statistics of all the variables, since it was not added before due to the fact that we ignore all csv files (fyi @gabrieleborg).

gabrieleborg:
As per discussion merging and closing this branch

Issue #15, opened 2020-03-18
Add alternative rents data


santiagohermo:
**[Goal]**
Add alternative sources of rents data to the build.

**[Notes]**
1. The main alternative source we discussed is [Fair Market Rents](https://www.huduser.gov/portal/datasets/fmr.html).
2. We should explore and add other sources of rents data that we think can be useful in the future.

gabrieleborg:
First pass: added Small Area Fair Market Rent Data as a dofile in base so far **not added to make.py**

The SAFMR data are annually and covers a fair number of zipcode (zcta).  Specifically, 

> The metropolitan areas identified to use ZIP code-based FMRs are those metropolitan areas with both significant voucher concentration challenges and market conditions where establishing FMRs by ZIP code areas has the potential to significantly increase opportunities for voucher families. The use of SAFMRs is expected to give HCV tenants access to areas of high opportunity and lower poverty areas by providing a subsidy that is adequate to cover rents in those areas, thereby reducing the number of voucher families that reside in areas of high poverty concentration. This page contains information to support the implementation of SAFMRs.

**important**: data from different years had different geographic identifiers. Still need to find the proper key variables.

Issue #24, opened 2020-04-03
Check data construction for geo codes, balance panel, fix event-study methodology


gabrieleborg:
**[Goal]**
As per April 01 2020 meeting, address the remaining methodological issues regarding the panel construction. In particular, make sure that our baseline panel(s) are composed of a set of fixed geo codes over time, such that for each geo code we have certainty of the county and state that it belongs so that we can assign information on MW changes.  Then re run analysis re-thinking event study methodology. 

**[Steps]**
1. Explore why there were some zipcdodes without MW change data as pointed out [here](https://github.com/diegogentilepassaro/min_wage_rent/pull/23#discussion_r403771596). 
2. Reassess the geo unit assignment in general. Consider mapping zipcodes to census tracts first if that helps. 
3. Make sure that each zipcode has assigned at least one and only one county. Each county then will have assigned a unique state.  Some zipcodes will belong to a MSA, the rest should be either in small places, or rural. We should probably find data on that distinction. 
4.  Create balanced panel(s) by actively selecting zipcodes with a populated enough time-series (e.g. all zipcodes with a fully populated time-series starting in 2012).
5. Re run analysis for baseline panel(s) re thinking through our event study methodology.

gabrieleborg:
Update: 

In commit 71287027e94651c6dc7ee5cc7b700f03327746d0 on related branch I address the issue of events too close to the dataset boundary (conditional an event window size) by _dropping_ events that are too close to it (see analysis/event_study/code/preclean_event_studies.do). 

This however made me think more broadly about our methodology as implemented in the present form: 

- There is still a confounding event effect in the post-dummies for the new last event: e.g. If I if I ignore Jan 2019, then the Jan 2018 event will have a confounding effect in the t+12 dummy)

- In general,  I suspect yearly changes in min wage are responsible for those not so smooth pre-trend as well as little jumps: e.g. if you look at the Jan 2018 pre-trend, that's technically the Jan 2017 post-trend as well (and indeed you see a little jump and an increase before stabilization around 6 months in).

- The best solution for last event maybe it's to restrict the event window to 11 months, so we don't capture the next or previous event.

[...]

PR #28, opened 2020-04-27
PR for #15:  add alternative Rent data

gabrieleborg:
@diegogentilepassaro  @santiagohermo 

For inital comments on Fair Market Rents data, please see https://github.com/diegogentilepassaro/min_wage_rent/issues/15#issuecomment-618434389. Respect to the first pass, I cleaned a bit the building procedure. Now the dofile `base/code/alt_rents.do`  appends all yearly Small Area Fair Market Rents by zipcode-cbsa, and returns a panel in `base/output/`.  At the moment, this is **not** run from the make.py script, so to test it you can simply run it independently. 

Small Area Fair Market Rents are Fair Market Rents calculated at the zipcode-level starting in 2016 for 24 metropolitan areas (cbsa) plus other non metropolitan areas that decided to adopt such measure voluntarily. In addition, HUD provides backward safmr starting in 2012/2013, which are computed for years until 2016 (even they didn't actually exist) using the formula then adopted.  

The mapping zipcode-cbsa is not so easy however: starting in 2018, new HUD Area Code were created by splitting some of the existing cbsa. The final dataset hence contains the name of the cbsa, which together with zipcode was the de-facto unique identifier untill 2017, and then a column named _hudcode_. This last is generally the same as the cbsa identifier (up to a starting character in front) for those geo-units that werent split, but it has changed for some zipcode. We can have therefore: 

1. the same zipcode in two different cbsa (even before 2018)

2. the same zipcode in the same cbsa until 2017, and then appearing in two different hudcode areas, one being the original cbsa. 

I couldn't decide the best way to deal with such situations, so I temporarily filled backwards the hudcode variable. In case whether there were two different hudcode in the post 2018 period, I backward filled the variable by taking the smaller hudcode as the main one. This is done creating a temporary variable called temphud. 

I then saved the data set using cbsa-temphud-zipcode-year as key.


diegogentilepassaro:
Thanks @gabrieleborg! I will review this tomorrow 👯

diegogentilepassaro:
I've started looking at this tonight actually! Some initial questions before I check the code:

1. I don't understand why we are not running it from make.py? Seems to me that it should be just changing a line from the template, but Is there something I am missing? 

2. The panel that we created goes from which date to which date?

3. I am not sure if I understand the issue with the zipcodes exactly (tomorrow I will think about it and maybe ask you). However, we want a panel keyed by the zipcode year. We don't care about the mapping with CBSA because we will have our own geocodes crosswalk soon. Does that make sense?  Seems to me that you just need to make sure that there are no duplicate zipcode in any given year. Whenever there are, keep the one that has more population within a CBSA or something like that.

gabrieleborg:
Thanks @diegogentilepassaro  for starting reviewing it. 

> I don't understand why we are not running it from make.py? Seems to me that it should be just changing a line from the template, but Is there something I am missing?

I didn't add it to make.py because you guys said to just get the data, organize it and store it. I thought we didn't want to run this as well all the time unless we were going to use it, but yeah it's just one line so we can add it if you want!

> The panel that we created goes from which date to which date?

The panel starts in 2013, the year selected for the rent series as well. As I wrote above, the Small Area Fair Market Rent started being utilized in 2016, but for informative purposes they uploaded the _hypothetical_ rate back to 2012.  These are the rate calculated with those years' information. 

> I am not sure if I understand the issue with the zipcodes exactly (tomorrow I will think about it and maybe ask you). However, we want a panel keyed by the zipcode year. We don't care about the mapping with CBSA because we will have our own geocodes crosswalk soon. Does that make sense? Seems to me that you just need to make sure that there are no duplicate zipcode in any given year. Whenever there are, keep the one that has more population within a CBSA or something like that.

Yes the main issue so far is that as we know the zcta - cbsa mapping is not 1:1. I can implement the same simplification we adopted with the Zillow data, but I was waiting to see what you guys do when reviewing the geography construction, as we may end up using another procedure? You let me know!

diegogentilepassaro:
Thanks for the answers @gabrieleborg! 

As for 1., as a matter of principle every time that we write code it has to live in the repo. If all you were going t do was to add raw data then it may or may not live in the repo depending on whether we care about people being able to reproduce how we obtain that data or not, and whether the data is obtained through code or not.

2. I am still confused. You say that the panel starts at 2013, but that they have data also for 2012. We should use the data from 2012 then! Also, a new related question: is the data varying at the year level?

3. I don't understand how this matters in this case. Say that a zipcode is such that it belongs to 2 different CBSA's. Do they have different values of the measure of interest or they have the same? If they have the same then we don't even have an issue and we don't care abut the mapping being 1:1 because we are interested in zipcodes and not in CBSA's. If they don't then, I think that the wisest thing to do for now is to tag those observations with a duplicate dummy. Then key the data by zipcode-cbsa-year that for each observation contains the rent measure of interest and a dummy for whether the zipcode-year belongs to more than 1 cbsa or not.

Let me know if this makes sense! Once you implement these changes I go ahead and review them.

gabrieleborg:
Hey @diegogentilepassaro ! 

I followed your comments and implemented the following changes: 

1. The code now is executed by main.py 

2. I added the year 2012 

3. I simplified the dataset: now the final output is a zipcode-year (yes, we don't have months here) dataset. The main problem here was that a given zipcode could be in one MSA and 2 counties, or in 1 counties and 2 MSA. I hence downloaded all zipcode-county and zipcode-msa crosswalks (note that these are USPS zipcodes, not Census zcta! this must be addressed in #24), since there are different crosswalks for every year (again, we may want to do that for the main dataset as well). Then I implemented the following procedure: 

-  First, I assign each zipcode to the county where the majority of residents live. This has been done to be consistent with what we have done so far. For reasons I couldn't figure out yet, there were few instances where the zipcode-county pair had no match in the crosswalk file. I hence selected the first pair.  

- Second, I then collapse each zipcode-msa using the zipcode-msa crosswalk. Also here, there are cases where the zipcode-msa pair wasn't present in the crosswalk. This means that I have to allocate some weight before collapsing. I therefore use the following rule: If, for a given zipcode, the non-missing weights sum up to one, I just use those pairs. If on the other hand there is weight missing, I re-distribute all the missing weight equally to all the zipcode-msa pair that originally didn't have it. See this example:
image
Zipcode 66547 has two zipcode-msa pair obs. When I merge the crosswalk however one of the pair gets a weight of 1. In that case, the final weights (new_wgt) will reflect that. On the other hand, zipcode 66834 got only a weight of .889 assigned. I therefore redistribute the remaning weight (missing_wgt) evenly across zipcode-msa obs with missing weight (n_mis). In this case all the missing weight goes to the only zipcode-msa pair. 

We finally ends up with a clean dataset, here is a snippet of my data file manifest: 
image

Issue #16, opened 2020-03-18
Add Wages and Employment data


santiagohermo:
**[Goal]**
Add data on wages and employment at the most disaggregated level.

**[Notes]**
1. We have in mind to use the [Quarterly Census of Employment and Wages](https://www.bls.gov/cew/) as main source.
2. We should consider adding data from the census for available years as well.

diegogentilepassaro:
I've started working in this task. The data should be downloaded form [this website](https://www.bls.gov/cew/downloadable-data-files.htm) and we should use the NAICS-Based Data Files. That website also contains the links to access data guides and dictionaries. 

First some comments on things that we have to decide:

- The data is available in several formats. I went over all of them and the useful ones are:

    1. County High-Level: for each year download comes a file for each quarter within that year. For each file we have all counties and all MSAs separated into 12 very broad industry codes. Contains employment at the monthly-county-industry level, number of establishments  at the quarter-county-industry, and average weekly wage within each quarter-county-industry. Does not contain geo unit FIPS code. 

    2. CSVs by area quarterly: for each year download we have one file per geo unit (state, county or MSA) containing FIPS code and unit name in the name of each file.  Within each file there is information on all quarters for the same variables as in 1. but there is also some extra variables (potentially useless) and (crucially) area FIPS code for each geo unit. 
![Screen Shot 2020-04-03 at 11 56 08 AM](https://user-images.githubusercontent.com/20582494/78381309-2f2fc400-75a3-11ea-804f-53f772b8ebcf.png)

   3. CSVs by industry quarterly: same as 2. but within each year download there is a file by detailed industry code (hundreds of categories) containing information on that industry code activities for all quarters for each county and MSA.  Some examples below. 
![Screen Shot 2020-04-03 at 12 05 09 PM](https://user-images.githubusercontent.com/20582494/78381464-6605da00-75a3-11ea-8ef0-0424b7f0a1d2.png)

- My hunch is that we need to create a crosswalk between county names and area FIPS codes using files in 2. or 3. and then we will probably be OK using files from 1. but adding the area FIPS codes through the crosswalk. 12 industry codes seems good enough although we can use files from 3. for some particular industries that we think relevant (like all that have an average weekly wage of less than something).  Something to discuss but we probably will need to upload to our data files from 1. for sure and potentially some from 3.

santiagohermo:
Thanks for that summary @diegogentilepassaro! Those data look great.
Looking forward to talk about them in the code.

[...]

PR #29, opened 2020-05-11
PR for #24

diegogentilepassaro:


diegogentilepassaro:
@gabrieleborg and @santiagohermo, this branch is now ready for review. I've spent the morning streamlining code, thinking about how to present a lot of information in the bootstrap table, and how to really show how robust the results are. The output is the following table (which I think is really cool! :D):

Screen Shot 2020-05-12 at 12 37 15 PM

Basically, I've added information on the number of events, some summary stats on size of the events, and I have run the bootstrapped passthrough for different size assumptions and different event definitions. Results are very neat and incredibly robust.

Please review these and let's merge to master. I will have lunch and start working on the #16!

gabrieleborg:
@diegogentilepassaro @santiagohermo Just finished running everything...it works perfectly! Thanks Diego for the effort and the results!

Looking at this I think it'd be interesting (in the future) to explore heterogeneity not only in terms of where the minimum wagers are, but also in terms of house sizes maybe? Anyway great job! Let's merge this asap!

gabrieleborg:
It doesn't let me review though..

analysis/bootstrap_passthrough/code/bootstrap.do
```diff
@@ -0,0 +1,86 @@
...
+program thing_to_bootstrap, rclass
+    syntax, event_var(str)
+	
+    reghdfe medrentpricepsqft_sfcc ib24.last_`event_var'_rel_months24, nocons ///
+		absorb(zipcode calendar_month year_month)
+	local sum_coeffs = 0
+	forval i = 25(1)49 {
+        local sum_coeffs = `sum_coeffs' + _b[`i'.last_`event_var'_rel_months24]
+		local effect_per_sqft = `sum_coeffs'/25
+	}
```
gabrieleborg, line 61: I just wanted to mention that the sfcc series include _single family_ but also _condos_ as type of houses. In the future we could replicate this exercise for _single family_ only as they are among the series with the most number of zipcode as reported in https://github.com/diegogentilepassaro/min_wage_rent/issues/13#issuecomment-601911500

diegogentilepassaro:
@gabrieleborg exploring heterogeneity in terms of something like house size is a good idea. House size is probably highly correlated with wage income, but probably there is huge variation depending on the land availability (as in places with more land available supply can respond more and prices should repond less to income changes). Therefore, I would think that exploring heterogeneity in terms of free land available "a là Diamond" is an interesting avenue! 

Sorry that I didn't requested the reviews, now you should be able to review :D

analysis/bootstrap_passthrough/code/bootstrap.do
diegogentilepassaro: That is worth doing for sure! So far I have chose medpricerentpsqft_sfcc as baseline but maybe we can change to only _sf. We should also try out soon different number of bedrooms and things like that.

diegogentilepassaro:
Great @gabrieleborg! I am working on the merge now but will make a fresh run of everything to be 100% sure!

santiagohermo:
You move quite fast guys! Haha. I'll make a quick review of the code now to make sure I understand what we are doing and keep up with our implementation.

derived/code/addMinWage.R
```diff
@@ -7,7 +7,6 @@ load_packages(c('tidyverse', 'data.table', 'matrixStats'))
 
 main <- function(){
    datadir   <- '../../base/output/'
-   outputdir <- "../../drive/derived_large/output/"
    tempdir   <- "../temp/"
    log_file  <- "../output/data_file_manifest.log"
```
santiagohermo, line 11: We can delete this line if we are not saving a logfile. 

You think the code is fast enough it's not worth it to keep this file in the drive? Maybe it'll make sense after we add amenities and more variables?

derived/code/create_final_panels.do
```diff
@@ -0,0 +1,68 @@
...
+	    medrentpricepsqft_sfcc medrentpricepsqft_mfr5plus medrentpricepsqft_2br {
+	    create_baseline_panel, var(`var') start_date(01jan2013) end_date(12dec2019)
+	}
+	use "../temp/baseline_medrentprice_sfcc.dta", clear
+	foreach var in medrentprice_mfr5plus medrentprice_2br ///
+	    medrentpricepsqft_sfcc medrentpricepsqft_mfr5plus medrentpricepsqft_2br {
+		merge 1:1 zipcode year_month using "../temp/baseline_`var'.dta", ///
+		    nogen keep(1 3)
+		}
+	save_data "../../drive/derived_large/output/baseline_rent_panel.dta", key(zipcode year_month) replace
```
santiagohermo, line 17: I think the log file of this files should live in `./derived/output/`, right?

derived/code/create_final_panels.do
```diff
@@ -0,0 +1,68 @@
...
+	keep zipcode
+	save_data "../temp/`var'_nonmiss_zipcodes.dta", replace ///
+	    key(zipcode) log(none)
+	restore
+	
+	merge m:1 zipcode using "../temp/`var'_nonmiss_zipcodes.dta", ///
+	    keep(3) nogen
+		
+	xtset zipcode year_month
+	assert r(balanced) == "strongly balanced"
```
santiagohermo, line 62: This is a very nice trick! :clap:

derived/code/clean_gen_label.do
```diff
@@ -0,0 +1,62 @@
...
+        bysort event_month`var_type'_id: gen months_until`var_type' = _N - months_since`var_type'
+
+        bysort event_month`var_type'_id: replace months_until`var_type' = 0 if _N == months_until`var_type'
+
+        drop event_month`var_type'_id event_month`var_type'        
+    }
+	
+    gen sal_mw_event = (dactual_mw >= 0.5)
+	gen mw_event025 = (dactual_mw >= 0.25)
+	gen mw_event075 = (dactual_mw >= 0.75)
```
santiagohermo, line 58: Thought: We should try a definition based on % change with respect to old MW and with respect to average wage (qcew!)

derived/code/addMinWage.R
diegogentilepassaro: Good eye @santiagohermo! will remove the log file line in #16.

I think that now the code is fast and the data is not as large. However, let's keep as it is for now so that we can reassess once we have the amenities data onboard.

derived/code/create_final_panels.do
diegogentilepassaro: That's right! I will fix that in #16 as well!

derived/code/create_final_panels.do
diegogentilepassaro: :D

derived/code/clean_gen_label.do
diegogentilepassaro: That is a good idea. Let's plan how do robustness based on that too, although we will have to change the definition of what we call passthrough. We would probably have to measure the income increase as a share of old income, and rent increase as a share of old income or something like that.

derived/code/addMinWage.R
santiagohermo: Coool!

derived/code/create_final_panels.do
santiagohermo: Learning how to code from the masters here haha

derived/code/clean_gen_label.do
santiagohermo: Yeah that sounds good. Hopefully we get the same pass-through and then we celebrate.

Issue #31, opened 2020-05-30
Add demographics data


gabrieleborg:
**Goal**

We want to add socio-demographic characteristics so to perform heterogeneity analysis over the MW impact on the housing market.

**Steps**

1.  Download SES data from 2010 census
2. Add SES data to main dataset

gabrieleborg:
First pass of demographics @diegogentilepassaro  @santiagohermo!

I added demographics to the pipeline in a two-step procedure: 

1. I created `base/demographic` to host the code assembling the different tables from the census, and creating the demographic variables from the raw ones (e.g. the share of college graduates starting from the raw counts of residents with various educational attainments). The output is a simple USPS zip-code level csv file (see data manifest [here](https://github.com/diegogentilepassaro/min_wage_rent/tree/issue31-AddDemographics/base/demographics/output)). 

2. I added the demographics to the final, baseline datasets created with `derived/zipcode_month/code/create_final_panels.do`.  This is done fore the moment by simply setting a local switch to "yes": once the local is on, demographics are added to `drive/derived_large/output/baseline_rent_panel.dta` and  `drive/derived_large/output/baseline_listing_panel.dta` before saving them. I implemented it here as I didn't want to carry around this demographics too much in the dta building process since the data is already large enough. In this way we are adding info only to those zipcodes we end up using and we keep the data as light as possible.  If you believe we can implement it in other locations let me know! 

Right now this task should be basically over. I saw that @diegogentilepassaro opened #33 but he didn't mention heterogeneity analysis there. I will experiment a bit tomorrow, so let me know if you prefer to continue on this branch/issue or whether we shall do it on #33 !

diegogentilepassaro:
Hola @gabrieleborg!

Quick question, are we going to use this issue to explore the heterogeneity by, for example, zipcode median income? Or should we move this to PR and do that in a new issue once everything is reviewed?

gabrieleborg:
Hey guys! @diegogentilepassaro @santiagohermo 

I was about to answer that! In general I was thinking to PR this and open another task, but if you both want to do it here I have no problem :) 

[...]

Issue #27, opened 2020-04-21
Amenity Index


gabrieleborg:
**Goal**

Understand  how to build , and create, an index for amenities that allows to capture local changes in response to MW treatments so to pin down the endogenous sorting of people across zipcodes. This measure should account for the fact that the event-study specification is already rich of FE controlling for local dynamics. Clarify the role GPS location data can play in mapping zipcode residents' preferences for various type of amenities, and how this can contribute to the construction of the aforementioned index. 

**Steps**

1. Literature Review:

   -   How amenities are measured, and at which geographic and temporal level. 
   -   How spatial patterns can identify people preferences

2. Identify simple/reproducible facts that can test the main assumptions to be validated

3. Start investigating (also on a sample) the potential dynamics it is possible to quantify

gabrieleborg:
Morning Guys! @diegogentilepassaro @santiagohermo 

Points 1 and 2 of the Steps list are being investigated both with you during our calls, and  individually I am keep reading papers and ideas. Since we need a boost for this issue I am anticipating step 3, which ideally will inform us also on the previous ones in a feedback loop.

I started creating the code to process the GPS data in order to create some initial potential measure for amenities. The data comes in Points-of-interest (POI)-monthly pattern cross-sections, that joined together create the monthly panel. I put the raw data in our Gdrive, in the subfolder `SafeGraph` along with all the complementary files that are needed (e.g. each POI info such as the industry categorization, that comes in a different file). 

As a first pass, I am focusing on computing the median income of visitors for each zipcode's _restaurants and eating places_ NAICS 4-digits categorization ([link](https://www.naics.com/code-search/?naicstrms=7225)). This includes therefore, bars, cafeterias and fast food as well. 
To do this, I organized the code in 2 scripts living in `base/code` that take care of the following steps: 

1. For each monthly file, extract the raw data, remove unused variables, normalize visits and compute the median income of each POI visitors. The visit normalization is done in order to account for sample bias (i.e. certain geographies may be over-represented) and temporal growth of the sample of devices tracked. I followed what has been identified as one of the best practice to normalize SafeGraph Data in the community of researchers using it, but this is not the only way to do it and we could talk about it. Specifically:  

    - for every POI-month, we observe number of visits and number of unique visitors. The first assumptions is to assign to each unique visitor the same number of visits (e.g. if place x has 100 visits and 75 unique visitors, we'll have that each visitor has done on average 1.33 visits). 

[...]

PR #32, opened 2020-06-01
PR for #16 add wages employment

diegogentilepassaro:
Alright @santiagohermo and @gabrieleborg, finally I am sending #16 to PR, after a small hiatus moving and traveling!

The original aim of that issue was to build some useful county-quarter and/or county-quarter-industry panel with wages and employment from the QCEW data that could be merged into our baseline derived panels. That was somehow achieved, but we also have used this task to sprint for having a draft for May 15, and to get there several important changes have occurred. 

The things that need a thorough review are the following:

1. The rearrangement of the base and derived directories into several directories that run independently. This includes making sure that the drive is updated with the new derived_large files. Also, let's think about wether is still convenient to keep that structure given that the datasets we have are very close to final in terms of the size they will have. 

2. The fix of the geocodes matches through adding a USPS zipcode to ZCTA mapping (Note that the zipcode in our baseline panels is the USPS zipcode as that is the id of the zillow data). This steps happen both in base and derived and both directories need to be review carefully to understand what's going on. We need to make sure that each zipcode from the Zillow data is identified correctly as a ZCTA that can be mapped to county, MSA, and state so that our MW changes are well defined for all units. One way to do this is to each of you check 10 random different zipcodes. Let's create a google sheets for doing this and let's make sure that all makes sense. 

3. The creation of a very basic county-quarter panel that rolls the zipcode-month data using the quantity of houses in each zipcode as weights. 

4. The re-do of the balancing of the panels at the baseline panels based on a balancing date that is different to the starting date of the panel.  That is for a given balancing dates, we choose all zipcodes that were available in the zillow data for that date, and balance the panel backwards and forward. 

5. Check the new analysis directories, and make sure that they are all using the same data, the same event-study code, and the same specification where appropriate. Let's also create a spreadsheet specifying inputs (datasets) and outputs (figures or tables) of each analysis directories so that we all are in the same page for the future and so that we organize our outputs for the coming draft. As a final mission for this point, let's check that the objects that live in the current draft are built in the latest version on this issue and also list in the spreadsheet which figure are they in the draft. 

Another separate point for review that I spotted 

- I couldn't figure it out exactly, but playing with the data I discovered that there is something weird about some MW changes that go negative. For example, the city of San Francisco has a change in MW that affects some zipcodes in our data. However, in our data the MW change was only effective for a month and then got back to the old MW level. That is not the case. What I did for now is to set all MW changes that are negative to 0 for now at the derived stage, but we should fix that as part of this PR too. Let's start by figuring out what happens with San Francisco first: if it is a data mistake or code and move from there.

A final point that would be nice to think and discuss about: 

- It would be nice to compile the .tex file in code within the draft directory. Let's also open a discussion about how to do that! 

**Let's try to make this issue a priority so that we merge to master and we have a stable code version that builds the current draft, so that we can move on to the next outstanding issues.**


santiagohermo:
Thanks @diegogentilepassaro! Those comments are very helpful. I will start to work on this PR now.

About your last point. I noticed that there is an [official `gslab_make` package for python 3](https://github.com/gslab-econ/gslab_make). After this is a done, we can create a new issue to update our make files and automate the building of the draft. (Unfortunately, the gslab_make package only builds lyx. We'll have to figure out how to build latex from the shell. It shouldn't be a problem)

santiagohermo:
I think we should order a bit our raw data, adding relevant documentation and small README files (e.g., I found the `fhfa` data and I have no idea how it got there, and why it's not in the GDrive). What do you think of this structure @diegogentilepassaro @gabrieleborg ?

repo
-- raw
------- CaseShiller
-------------- docs
-------------- readme.txt
------- ...
------- qcew
-------------- code
-------------- docs
-------------- readme.txt
------- ...
-- drive/raw_data
------- CaseShiller
-------------- files
------- ...
------- qcew
-------------- files
------- ...

The repo should contain documentation and code to download files, the drive/raw_data folder should contain only the files. I will order the qcew folder like this.

santiagohermo:
I think I fixed the issue with San Francisco and other local mw @diegogentilepassaro . In base\substate_mw.do we were creating the daily minimum wages with `tsfill`
https://github.com/diegogentilepassaro/min_wage_rent/blob/97440e7b513fedff2ed17abaadd1488eb0adb157/base/zillow_min_wage/code/substate_mw.do#L110-L111

`tsfill` fills until the last data available. In San Francisco, the last data was 2019-7-1, so after that it was assigning the state mw to the are.

To solve this I made sure that every place has a row some moment after 2019-12-31 with their current local mw (if they hadn't one, I added for "2020-1-1" with the current local mw). So it shouldn't be an issue in the future.

diegogentilepassaro:
> I think we should order a bit our raw data, adding relevant documentation and small README files (e.g., I found the `fhfa` data and I have no idea how it got there, and why it's not in the GDrive). What do you think of this structure @diegogentilepassaro @gabrieleborg ?
> 
> repo
> -- raw
> ------- CaseShiller
> -------------- docs
> -------------- readme.txt
> ------- ...
> ------- qcew
> -------------- code
> -------------- docs
> -------------- readme.txt
> ------- ...
> -- drive/raw_data
> ------- CaseShiller
> -------------- files
> ------- ...
> ------- qcew
> -------------- files
> ------- ...
> 
> The repo should contain documentation and code to download files, the drive/raw_data folder should contain only the files. I will order the qcew folder like this.

Reorganizing raw was a great adiea!

Also @santiagohermo, the fix for the San Francisco issue looks good!

diegogentilepassaro:
@gabrieleborg and @santiagohermo, one last thing that you guys should think about when doing this PR that I forgot to mention is something that we talked in our last meeting. We should spot in our code where we can add identifiers for each type of MW change (city, county, state, federal) so that soon enough we can play around by creating some table that shows the differences in results across the different types of events, and so that we get a sense of which variation is key (probably will be the state, but let's see!). Knowing this will also inform our choice of clustering. 

Probably the best way to go is to post a comment on where you think this should be done, and after discussion we can implement.

santiagohermo:
[That](https://github.com/diegogentilepassaro/min_wage_rent/pull/32#issuecomment-638227655)'s a good point @diegogentilepassaro. 

For the zipcode_month panel we are creating the mw vars in [addMinWage.R](https://github.com/diegogentilepassaro/min_wage_rent/blob/78be38687acf420172dd4c8f5231ca2685e30e6d/derived/zipcode_month/code/addMinWage.R#L33). Maybe we can add dummies there, link 'state_mw_ind', 'county_mw_ind', 'local_mw_ind' ? @gabrieleborg created that file so maybe he knows best here.

diegogentilepassaro:
That sounds like a good idea @santiagohermo! Let's see what @gabrieleborg says :D

santiagohermo:
I have reviewed `raw`, `base/qcew`, `base/zillow_min_wage` and `derived`. What a great pipeline we have built @diegogentilepassaro @gabrieleborg! 

I did some reorganizing of the code and style changes mostly, with some important things I mention below. Also, I updated Google Drive linked to those folders accordingly (I recommend we only change GDrive after a pull is finished, so that it is always up-to-date with master).

I will start working here again tomorrow afternoon!

#### Noteworthy Changes

In `raw`:
- I added some readme files [here](https://github.com/diegogentilepassaro/min_wage_rent/tree/83d5c82f58597ea1e3cad08f152fb20f1e769453/raw). Most of them need completion. 
- Regarding the QCEW build
   - I changed the download of the QCEW only to County-High level (since this is the only files we are using). For this I just commented out some lines of code. The [GDrive](https://drive.google.com/drive/u/1/folders/1mfXcqWVt0wtMAK2xowuvri38Y77gS6uy) matches this build
   - I added the year 2009. This is just in case we won't to do something with first differences, it is good to have a year before of data!
- I moved crosswalks to [`raw/crosswalk/`](https://github.com/diegogentilepassaro/min_wage_rent/tree/496639ee0c514f183edd08892b1ead771f164d37/raw/crosswalk)

In `base`:
- Mostly style changes
- Importantly, they bring in [the fix](https://github.com/diegogentilepassaro/min_wage_rent/pull/32#issuecomment-637901618) in some substate mw changes
- Per [Diego's comment](https://github.com/diegogentilepassaro/min_wage_rent/pull/32#issuecomment-638227655), we want to add indicators for where the mw change is coming from in `base/zillow_min_wage/code/substate_mw.do`

In `derived`:
- Again mostly style changes that bring in the fix of some substate mw changes

#### Questions
1. We should probably bring the file `GDrive/DownloadCensusData.R` to the `raw` folder in the repo.

2. In GDrive there are many files correspoding to the build of amenities in #27 that won't match master after this PR is merged. Since it seems we won't be adding amenities anytime soon, I would delete them (expect the `raw_data`, of course). But I don't feel strongly about it so we can keep them if you want.

3. About the files in the [GDrive/raw_data/hud_Xwalks](https://drive.google.com/drive/u/1/folders/1YabBwqgwMISrJgisy4YCcRu1P3Dn0HAV). Are we using these? Maybe I missed that bit of code. In case we are using them, we need to bring them to the [`raw/crosswalk/`](https://github.com/diegogentilepassaro/min_wage_rent/tree/496639ee0c514f183edd08892b1ead771f164d37/raw/crosswalk)

4. About `base/alt_rents`. Do we want to bring these files to the build in this PR? I haven't reviewed this

5. About [`derived/zipcode_month/code/clean_gen_label.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/78be38687acf420172dd4c8f5231ca2685e30e6d/derived/zipcode_month/code/clean_gen_label.do). In particular, these lines:
https://github.com/diegogentilepassaro/min_wage_rent/blob/78be38687acf420172dd4c8f5231ca2685e30e6d/derived/zipcode_month/code/clean_gen_label.do#L25-L26

I tried to change them to `assert`, but it failed. So I guess we still have missings. Is this because the matching from zipcodes to zctas is not perfect?

6. In `derived/county_quarter/code/preclean_qcew.do` we are splitting counties based on the period 2010q1. Shouldn't we do this based on the full 2009 instead (since it's now in the data)?

diegogentilepassaro:
Thanks for your great work @santiagohermo!

As for adding the indicators for the type of MW change in [`derived/zipcode_month/code/addMinWage.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/78be38687acf420172dd4c8f5231ca2685e30e6d/derived/zipcode_month/code/addMinWage.R) let's wait for @gabrieleborg thoughts so that we decide in which lines it would make sense to add this. 

As for your questions:

1. I agree! We should probably do this as part of #31. @gabrieleborg I will make a note there. 

2. Keeping only the raw_data seems like the way to go to me, but let's see what @gabrieleborg says!.

3. Thanks for spotting this! In the end we are not using them. We should delete it unless @gabrieleborg thinks that could be useful for the census data mapping. 

4. That is a very good question. I've tried to run that directory but it didn't go through. If @gabrieleborg agrees I think that it would be useful to bring to master so that we have more inputs to start writing a descriptive section. With the FMR data we can do some plot comparing to the Zillow data. With the census data we can create some table that presents the data that is now written in the draft in the data section.

5. Very good point. I am not sure what the issue is but I will look into it and write a comment about it. One quick question: your link points to master and not to the issue16 branch, are you sure that it fails in the branch? Which assert fails? Zipcodes or year_months? 

6. Not sure what you mean. The counties are divided based on the wages, and the wages data is at the quarter level, so what do you mean by the full 2009? An employment-weighted average of the quarterly data rolled to the year? or something else? Also, I like 2010 as a year because that's the year of the census, so we can compare with the wages rankings for each county that we will obtain from there when rolling the ZCTA's to the county.  

Let me know your thoughts!

gabrieleborg:
hey Guys! @diegogentilepassaro  @santiagohermo sorry for the delay. 

As Santi, for now I reviewed `raw`, `base/qcew`, `base/zillow_min_wage`, and `derived`. The re-organization looks fantastic! Here are my comments first and then answers: 

- In `raw`, I added a description for all datasets that were missing it. 

1. I deleted the `Gdrive/DownloadCensusData.R` as I ended up not using it. 

2. You refer to the two dataset in `base_large/output` ? Yes we can remove them (I'll do it), as once we have all updates in place and we'll eventually come back to that we'll surely have to adjust the construction of those indexes.

3. I am planning to use some in my task "add demographics". Let's remove the one we don't use once we review that task (hopefully soon). 

4. I reviewed `base/alt_rents` and it runs smooth and fast. That's weird @diegogentilepassaro. If you manage to double check it and make it work, let's bring it to PR!

5. I'll have a look tomorrow. 

6. I don't fully understand either. I think @santiagohermo  is making the point that, now that we also have 2009 data, it could be good to add those. As for the census, I don't think it's a big issue (John always minimizes the temporal variability of SES variables): if it brings us a considerable increase in power or precision fine, but I have no strong feeling about this. 

Tomorrow I'll finish my review of the whole repo. Good work!

santiagohermo:
Thanks for those thorough answers @diegogentilepassaro @gabrieleborg ! Brief reactions:

About raw. In some files it says "Complete Description". What that means is we need to write there what the fails in the indicated folder. We can do this as we go along.

1. Got it.

2. Sounds good Gabri! I'll let you remove those then. After this PR we should have the GDrive perfectly aligned with Google Drive 💪 .

3. Great! One you do that, remember to add the files to `raw/crosswalk/` and delete them from GDrive.

4. I agree in that it would be great to bring those data to the build. In my meeting today with Ken and Jesse they expressed some concern in that those data may have "endogenous measurement error", meaning (I think) that the composition of houses available for rent may change when the minimum wage changes. Using rents from an alternative source may alleviate those concerns.

5. Actually, my link points to a tree of the branch being reviewed here. Nothing fails there. What I meant is that I changed those lines the "drop if" to an "assert" and STATA stopped, thus signaling that yearmonth has missing. I think it would be nice at least to know how many these are and where they are coming from.

6. What I meant is that I would use data before the estimation panel starts to classify observations. But it's not a big deal. Also, I think it's useful to base our classification on the whole year as there may be seasonality in using only one quarter (e.g., employment in beaches will be probably low in the first quarter)

derived/county_quarter/code/preclean_qcew.do
```diff
@@ -0,0 +1,44 @@
...
+		
+		gen avg_employment_quarter = (employment_month1 + employment_month2 + employment_month3)/3
+		egen us_total_employment_quarter = sum(avg_employment_quarter) 
+		egen us_avg_week_wage_num = sum(avg_employment_quarter*avg_week_wage)
+		gen us_avg_week_wage = us_avg_week_wage_num/us_total_employment_quarter
+		
+		gen above_avg_us_weekly_wage = (avg_week_wage > us_avg_week_wage)
+
+		egen decile_avg_week_wage = xtile(avg_week_wage), n(10)
+
```
gabrieleborg, line 39: That's very nice as we can easily compare counties and how they are impacted! Thanks @diegogentilepassaro! As I mentioned to you on the phone, I checked and even though such classification are based on time-varying information (the avg_week_wage), they remain constant within each zipcode. This hence allows us, fo the future, to run heterogeneity analysis by simply interacting the relative time dummies with either the above/below dummy or the deciles, so we can gain in sample size and precision. At the same time, I still believe it's higher priority do this exercise at the zipcode level using census data as soon as we merge this branch.

PR #36, opened 2020-06-13
PR for #31 - add demographics

gabrieleborg:
@diegogentilepassaro @santiagohermo  following the discussion in #31 I am opening this PR to add demographics. All meaningful changes have been already documented  [here](https://github.com/diegogentilepassaro/min_wage_rent/issues/31#issuecomment-640339798) and [here](https://github.com/diegogentilepassaro/min_wage_rent/issues/31#issuecomment-643414223).


diegogentilepassaro:
Good morning @gabrieleborg!!! The addition of the demographics looks great. Just one quick comment about the flagging of the type of events: What you flagged is which MW law is binding in each zipcode but I believe that it would be easier going forward to flag instead each *event* (defined by \Delta actual_mw != 0). It is conceptually more clear and also I believe that flagging each moment has issues. For example, supose that at some period in time the bindig MW is the local one and later is the state one. We don't know from the flag if it is because the state one raised above the local, or because the local was abolished. Therefore we don't know the geographic level of the event that we are using for estimation, which is what readers will care about.

Does this makes sense? What do you think @santiagohermo ?

base/demographics/code/make_demographics_ziplevel.R
```diff
@@ -0,0 +1,203 @@
...
+      data[, 'housing_units2010':= IFC001]
+      
+      target_vars <- c('tract_fips', 'county_fips', 
+                       'pop2010', 
+                       'urb_share2010', 
+                       'white_share2010', 'black_share2010', 'hisp_share2010', 'asian_share2010', 'natam_share2010', 
+                       'child_share2010', 'teen_share2010', 'youngadult_share2010', 'adult_share2010', 'elder_share2010', 
+                       'housing_units2010')
+      
+      data <- data[, ..target_vars]
```
santiagohermo, line 109: This line is repeated under each if. It could go outside of all the conditionals instead, to avoid repetition.

derived/zipcode_month/code/create_final_panels.do
```diff
@@ -45,6 +57,10 @@ program main
 		merge 1:1 zipcode year_month using "`instub'/baseline_`var'.dta", 	///
 			nogen keep(1 3)
 	}
+	if "`add_demo'" == "yes" {
+		merge m:1 zipcode using `instub'/zip_ready.dta, nogen assert(1 2 3) keep(1 3)	
+	}
+
```
santiagohermo, line 63: In this file you are not adding the demographics to "zipcode_yearmonth_panel_all.dta". Just flagging that, if we want to use them for the simulation (which builds from that file), it would be useful to add them.

base/demographics/code/make_demographics_ziplevel.R
```diff
@@ -0,0 +1,203 @@
...
+      data[, c('child_share2010',      #0-14
+               'teen_share2010',       #15-21
+               'youngadult_share2010', #22-34
+               'adult_share2010',      #35-64
+               'elder_share2010') :=   #65-
+              list((H76003 + H76004 + H76005 + H76027 + H76028 + H76029) / H76001, 
+                   (H76006 + H76007 + H76008 + H76009 + H76030 + H76031 + H76032 + H76033) / H76001, 
+                   (H76010 + H76011 + H76012 + H76034 + H76035 + H76036) / H76001, 
+                   (H76013 + H76014 + H76015 + H76016 + H76017 + H76018 + H76019 + H76037 + H76038 + H76039 + H76040 + H76041 + H76042 + H76043) / H76001,
+                   (H76020 + H76021 + H76022 + H76023 + H76024 + H76025 + H76044 + H76045 + H76046 + H76047 + H76048 + H76049) / H76001)]
```
santiagohermo, line 97: pretty neat data.table notation 👏

base/demographics/code/make_demographics_ziplevel.R
```diff
@@ -0,0 +1,203 @@
...
+      y[, c('tract_fips', 'county_fips') := list(
+         as.numeric(paste0(
+            str_pad(STATEA, 2, pad = "0"),
+            str_pad(COUNTYA, 3, pad = "0"),
+            str_pad(TRACTA, 6, pad = "0"))),
+         as.numeric(paste0(str_pad(STATEA, 2, pad = "0"),
+                           str_pad(COUNTYA, 3, pad = "0"))))]
+
+      return(y)
+   }
```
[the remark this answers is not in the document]
gabrieleborg: In this way it's easier to experiment as you can run the script manually by itself from RStudio piece by piece. I have no strong preference about this!

base/demographics/code/make_demographics_ziplevel.R
gabrieleborg: You're right, thanks!

derived/zipcode_month/code/create_final_panels.do
gabrieleborg: Added.

base/demographics/code/make_demographics_ziplevel.R
```diff
@@ -0,0 +1,203 @@
...
+source("../../../lib/R/save_data.R")
+
+load_packages(c('tidyverse', 'data.table', 'tidycensus', 'bit64', 'readxl'))
+
+main <- function() {
+   
+   data_version <- "0043"
+   
+   datadir <- paste0("../../../drive/raw_data/census/tract/nhgis", data_version, "_csv/")
+   xwalkdir <- "../../../drive/raw_data/hud_Xwalks/"
```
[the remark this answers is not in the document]
gabrieleborg: I updated it!

gabrieleborg:
Thanks @diegogentilepassaro  @santiagohermo for the review. yeah I can add that :)

I update the code following some of Santi's suggestions, I will do this soon!

base/demographics/code/make_demographics_ziplevel.R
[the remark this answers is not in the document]
santiagohermo: I see. I have no preference either, just thinking about your code style. It's always nice to learn new tricks haha

base/demographics/code/make_demographics_ziplevel.R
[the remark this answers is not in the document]
santiagohermo: That looks good, thanks!

santiagohermo:
Thanks @gabrieleborg! On my side, after you add those indicators I think we are good to merge here.

gabrieleborg:
hey @diegogentilepassaro  and @santiagohermo , 

I added 4 dummy variable (fed_event, state_event, county_event and local_event )that turn on whenever there's a change in MW due to the associated geographic level. You can see the new piece of code in c64ec2f

Issue #41, opened 2020-09-08
Repo maintenance


santiagohermo:
After the summer we are starting to work again on this project. This task is a part of that process, in which we will:

- Drop things we won't be using in the future (event studies?)
- Improve code or organization of files if needed
- Set up [gslab_make with python 3](https://github.com/gslab-econ/gslab_make) (santi)

I will focus on this task this week, requesting your input when needed @gabrieleborg @diegogentilepassaro

santiagohermo:
Question @diegogentilepassaro @gabrieleborg. Do we want to keep the county-quarter analysis, or should I delete it?

gabrieleborg:
@santiagohermo mmm I think we could keep it for the moment, and delete it maybe later? I am saying this because I will have to send a draft soon to Matt and Emily, and if they ask us to substantiate a bit more why our analysis is superior to previous papers, we can readily put a couple of plot in Appendix.

santiagohermo:
Sounds good @gabrieleborg, thanks for the quick reply!

diegogentilepassaro:
I agree with keeping it too! We can probably improve the hay we’ve done at a later stage by adding directly Zillow data at the county-month and stress the importance of having small areas.

santiagohermo:
I updated `derived/zipcode_month` , but the data file manifest is showing that I added a lot of unintended variables (see [this diff](https://github.com/diegogentilepassaro/min_wage_rent/commit/04e4fb301e65ec0399762d0d2d8e7e225c3eeb33#diff-e7cccde3eb0401ed89f65ec7cd341bad)). I almost didn't change the code though, so I'm not sure why. Any thoughts on why this might be, @gabrieleborg?

santiagohermo:
Another question: I'm getting to updating `analysis`. It seems to me that we are creating `preclean`s dataset a lot there. What do you say if we add one folder (`analysis/estimating_panel` or something) to make the preclean only once, @diegogentilepassaro @gabrieleborg?

santiagohermo:
Per call, I'll wrap up this issue once we are working on the draft (to avoid conflict with existing branches)

gabrieleborg:
[...]

Issue #17, opened 2020-03-18
Simulation: Build fake dataset and test event studies


santiagohermo:
**[Overview]**
This issue is a follow-up of #11 and, in particular, a suggestion of Jesse Shapiro to understand how the event-study design in first difference would look like.

**[Goal]**
Understand how the first-difference event-study design works and should look like in case minimum wage changes actually increase rents.

**[Steps]**
1. Simulate rents, minimum wages, and house values data per geographic unit. Consider having two kinds of geographic units in which the effect of minimum wages would be different.
2. Try out a first-difference event-study design on the simulated data.

**[Notes]**
- House values should be computed as the present value of discounted rents (plus a random shock?)
- Think about anticipatory effects in house values.

diegogentilepassaro:
@santiagohermo I think that now we are ready to tackle this task. Now that the last event methodology is more reliable and we have smooth 24 window period event studies for the same type of housing for both rental and listing prices it seems that we can compare our simulations with our results.

For example, for the type of house SFCC, the dynamics of rent per square foot look like:
![Screen Shot 2020-04-18 at 10 17 31 AM](https://user-images.githubusercontent.com/20582494/79640125-d4c76380-815d-11ea-9f10-b5a8e4b967f6.png)
and for listing price it looks like:
![Screen Shot 2020-04-18 at 10 17 11 AM](https://user-images.githubusercontent.com/20582494/79640131-dc870800-815d-11ea-9c56-6228f3505649.png)

Is the second plot what a simulation of listing price based on PDV of rent payments will look like if the rent payment change is the one that we computed in the first plot? To answer that question we will need to assume something about the moment of the announcement of the MW change. Is what we see consistent with some reasonable assumption? 

Possible assumptions to simulate:

1. The exact dynamics of rent prices around MW changes is know ex-ante.

2. The change is only known to happen 12 or 24 months before. 

3. The change takes landlords by surprise. 

Let me know your thoughts and we can start playing around soon!

santiagohermo:
Per call today, in this task we should use the existing dataset and simulate only rents and house values. With the simulated data, we should 
- try out our "last" methodology to compare with our results, and 
- try out the first-differences design.

santiagohermo:
In this branch I will simulate a fake dataset to try our "last" event study specification. This dataset will also be useful to test other specifications in the future

santiagohermo:
[...]

PR #42, opened 2020-09-08
PR for #17: Simulation

santiagohermo:
Do you guys want to take a look here @diegogentilepassaro @gabrieleborg ? I wrapped up the simulation in #17, which

- Creates a dataset with simulated data (see [code](https://github.com/diegogentilepassaro/min_wage_rent/blob/12836f179fdc1e0c8bc3804aaed1820f843e9c37/derived/simulation/code/simulate.R) and [data_file_manifest.log](https://github.com/diegogentilepassaro/min_wage_rent/blob/12836f179fdc1e0c8bc3804aaed1820f843e9c37/derived/simulation/output/data_file_manifest.log))
- Runs event studies with those data (in [this](https://github.com/diegogentilepassaro/min_wage_rent/tree/12836f179fdc1e0c8bc3804aaed1820f843e9c37/analysis/simulation) folder)

We will likely drop the event studies in #41, but I think that the simulated dataset is worth keeping. 

The reviewer should make sure the [code](https://github.com/diegogentilepassaro/min_wage_rent/blob/12836f179fdc1e0c8bc3804aaed1820f843e9c37/derived/simulation/code/simulate.R) that builds the fake data makes sense. This is of secondary importance though, so I'd be ok with merging without reviewing


diegogentilepassaro:
Very nice code @santiagohermo! I think that having this fake dataset will be really useful in the near future when we compare our different panel data approaches.

I will go ahead and merge and we can complete the housing values part later!

diegogentilepassaro (commented):
It all looks good!

Issue #44, opened 2020-09-10
Basic benchmarking of effects


santiagohermo:
[GOAL]
The goal of this issue is to come up with a simple computation to benchmark the effect of the MW on rents that we are finding now.

[STEPS]
1. Come up with a simple framework to express our object of interest, d ln r/d ln MW, in terms of parameters (share of MW earners in zipcode, share of income spent in earnings by MW earners, etc)
2. Come up with numbers for those parameters and figure out what effect they imply
3. Compare with our estimates of the effect

santiagohermo:
Hey @diegogentilepassaro @gabrieleborg, I had an idea for step 1 as we talked today which I just implemented! (I had to do it now or I would have lost it!)

Specifically, I assumed a very simple equilibrium in a zipcode housing market and derived an expression for our object of interest in terms of

- the elasticity of housing demand to the minimum wage and to rents for MW households
- the elasticity of housing demand to rents for non-MW households
- the elasticity of housing supply
- the minimum wage to rent ratio
- the share of housing occupied by MW earners
- the share of housing occupied by non-MW earners

I compiled my reasoning in [this pdf](https://github.com/diegogentilepassaro/min_wage_rent/blob/2e3274445b16efb60f9f8954a237cf44188b830b/issue44-benchmark/benchmark.pdf). Let me know your thoughts and whether you think it makes sense! Once you are onboard, we can move on to plugging in some numbers and getting our benchmarks

gabrieleborg:
That's very cool @santiagohermo ! 

I think this is a great start. My comments below: 

1. Can you show me how you obtain (1) though? I couldn't reproduce your results, as I am not sure how to fully differentiate with respect to 2 variables. 

2. Regarding equation (2), we can get shares and MW/rent ratio directly from the data (hopefully), but how would you compute the elasticities we need to benchmark our estimate ?

santiagohermo:
Thanks @gabrieleborg! Quick answers:

1. I'll add the derivation as soon as I have some time (I have it in my notebook here). Then we can check whether I screwed up or not haha
2. Those elasticites are the hard part, yeah! I would take the value from some paper, e.g., Diamond (2016). We can also just plug in made-up numbers that we think make sense

santiagohermo:
Update: new pdf [here](https://github.com/diegogentilepassaro/min_wage_rent/blob/7cf8323041e40bd076ce407428626bb483343dc9/issue44-benchmark/benchmark.pdf).

[...]

PR #45, opened 2020-09-17
PR for #44: Benchmarking

santiagohermo:
I think we are in a good place to wrap #44 up. We can play a bit with the parameters as we go along. I think gabri already went through the notes, so I assigned you @diegogentilepassaro. It should be a really quick review.

I stored the `benchmark.pdf` note in `analysis\notes`. If you have a better idea about how to store it, I'm all ears.

Let me know if you have any comments/questions!


analysis/notes/benchmark.Rmarkdown
```diff
@@ -0,0 +1,137 @@
...
+  
+  return(numerator/denominator)
+}
+```
+
+Suppose that the share of MW is $\underline{s} = 0.3$, so that $\overline{s}=0.7$.
+
+Finally, take the following values for the elasticities:
+
+- $\underline{\beta} = 0.1$, so housing demand is inelastic to income
```
diegogentilepassaro, line 125: Housing demand of the MW HH.

analysis/notes/benchmark.Rmarkdown
```diff
@@ -0,0 +1,137 @@
...
+- $\underline{\gamma} = - 0.7$ and $\overline{\gamma} = - 0.5$, so MW HH are more sensible to
```
diegogentilepassaro, line 126: This makes sense, but where are these numbers coming from? I am sympathetic towards the intuition that MW HH should be more sensible to rents, that is easy to defend. But do we at least have a sense of what is the elasticity of housing demand to rents that other papers estimate for the overall population or for some low or high income groups? 
A more complicated alternative, would be to assume that all households consume 1 unit of housing and what they also chose is the quality. Then things would be a function of the elasticities of housing quality to prices for each group and presumably we can get those from Van Dijk.

analysis/notes/benchmark.Rmarkdown
```diff
@@ -0,0 +1,137 @@
...
+```
+
+Suppose that the share of MW is $\underline{s} = 0.3$, so that $\overline{s}=0.7$.
+
+Finally, take the following values for the elasticities:
+
+- $\underline{\beta} = 0.1$, so housing demand is inelastic to income
+- $\underline{\gamma} = - 0.7$ and $\overline{\gamma} = - 0.5$, so MW HH are more sensible to 
+rents
+- $k = 0.1$, similar to estimate of Diamond (2016, Table 5)
```
diegogentilepassaro, line 128: Very nice that we take that parameter from Diamond (2016).

diegogentilepassaro (changes requested):
One big comment: all the benchmarking exercise, including the notes, should live on its own analysis directory- i.e. /analysis/benchmarking-. 
Then, the preclean of the first difference will remain simple and easy, and the benchmarking preclean one can have the needed variables that we added here.

analysis/notes/benchmark.Rmarkdown
gabrieleborg: the gamma parameters, as well as the income elasticity beta, could be recovered from Albouy et al (2016). I saved it in `bibliography/housing_market/`. Here they show that indeed the price elasticity is approximately 2/3 in magnitude, so -.5 and -.7 for the 2 groups seem reasonable. As I previously anticipated to you though, they also estimate the income elasticity to be around .5, while we use a way more inelastic one (.1). You can find their results in tables 1, 2,3, and 4.

analysis/first_differences/code/benchmark_test.do
```diff
@@ -0,0 +1,181 @@
...
+	foreach win in 5 {
+	benchmark_plot_all2, depvar(ln_med_rent_psqft) w(`win') absorb(year_month zipcode) cluster(statefips) outstub(`outstub') ///
+						 beta_low(`beta_low') gamma_low(`gamma_low') gamma_hi(`gamma_hi') k(`k')
+	graph export `outstub'/benchmark_all2_w`win'_ziptrend_base.png, replace
+
+	benchmark_plot_wmean, depvar(ln_med_rent_psqft) w(`win') absorb(year_month zipcode) cluster(statefips) outstub(`outstub') ///
+						 beta_low(`beta_low') gamma_low(`gamma_low') gamma_hi(`gamma_hi') k(`k')
+	graph export `outstub'/benchmark_wmean2_w`win'_ziptrend_base.png, replace
+	
+	}
```
santiagohermo, line 29: Given that we are not using `alpha` anymore, it looks like we don't need two functions. Right?

santiagohermo (commented):
Thanks @gabrieleborg for implementing the benchmark plots! Overall the code makes sense to me. I made a few comments to simplify it a bit

analysis/notes/benchmark.Rmarkdown
santiagohermo: Implemented!

analysis/first_differences/code/benchmark_test.do
```diff
@@ -0,0 +1,181 @@
...
+	sum actual_mw if e(sample)
+	local mw_month = r(mean) * `week_hours' * 4.35 * `mw_workers'
+	local alpha = `mw_month' / `rent_month'
+	
+	sum mww_shrenter_all1 if e(sample)
+	local mww_share_rent1 = r(mean)
+	sum mww_shrenter_all2 if e(sample)
+	local mww_share_rent2 = r(mean)
+	sum mww_shrenter_wmean2 if e(sample)
+	local mww_share_rentm = r(mean)
```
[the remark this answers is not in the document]
gabrieleborg: Yes, agree. But we still need the second block, to compute the average share of MW workers in the estimation sample!

analysis/first_differences/code/benchmark_test.do
gabrieleborg: That's slightly different, as the 2 functions compute the benchmark using two difference assumption on how to identify MW workers (that are needed in the benchmark as they enter as share). The first one assumes that all MW households are 2-workers households, while the second assumes that 1- and 2-workers households are uniformly distributed across income bins for renters ,and so the share of MW households identified through the census accounts for that.

gabrieleborg:
@diegogentilepassaro @santiagohermo I have implemented the requested changes. Let me know how it looks now.

diegogentilepassaro:
@gabrieleborg you are good to go on my end! Let's wait for @santiagohermo :D

santiagohermo:
Things look good on my end too, thanks @gabrieleborg!

Issue #43, opened 2020-09-10
Try more dynamic panel data specifications


diegogentilepassaro:
[Goal]
Have a table and a figure that compares several panel data specifications. We have a nice figure already. Maybe we add to that one maybe we create new ones.

[Notes]
- The main new specification that we want to try is (4) in [this paper](https://www.nber.org/papers/w19262.pdf) but we may also try specifications in levels.
- Also take a look at Appendix G from [Dube et al paper](https://oup.silverchair-cdn.com/oup/backfile/Content_public/Journal/qje/134/3/10.1093_qje_qjz014/4/qjz014_supplemental_online_appendix.pdf?Expires=1602680524&Signature=JwJWQ~TSE6I~RO-jhdm512nW8KS-XcRPlSUdSs4RO5sqYnowmusX2kXKGeeAAD3Y-I~7IIHpH1O1RfEuWuv0SaDIxlsVxuZH6MC4uORChNWXsSlWMkY-0oY~ufgJaLzLYuOOz5qOk1yMmgx6cOlWIE0kGBDNphGCermnLCSLiE6wXv5lIGtl7BTxL7RRAulvmyfBYKrqOCeA-HkWB7hjYNoReh6IAjh-gyrs6U6a3Rm1kMENqamjc201iHcqv7zWiBEXoLLNgxldAkAVCyuP~v0dS-cXX8tO8Z5HZWfpPl2gFVntfdTx-kVmtZRrbkT4mvz0g74NvVV8LBg40iCUOQ__&Key-Pair-Id=APKAIE5G5CRDK6RD3PGA) on MW effects on employment.

diegogentilepassaro:
Alright @santiagohermo and @gabrieleborg 

The nice table that I have been working on is ready :D Results are pretty robust to anything and extremely stable across model specifications. I have created in the branch issue_43 a tex file within /draft called staged_output.tex that I hope that we use for staging output that is waiting to be written into the draft. You can take look at the PDF 
[here](https://github.com/diegogentilepassaro/min_wage_rent/files/5211823/staged_output.pdf). 

Let me know if there is anything you want to discuss!

santiagohermo:
Thanks @diegogentilepassaro! That table looks pretty good, it even lowers the standard errors of the main estimates! Question: should we add the sum to those tables?

gabrieleborg:
Thanks @diegogentilepassaro ! 
[...]

PR #48, opened 2020-09-27
PR for  #43 - dynamic panel data specs

gabrieleborg:
As per our last conversation, I am sending #43 to PR. 

I additionally tried to build some MW change-date specific FE with the idea to compare MW changes occurring at different places at the same time, but this resulted to be less intuitive than I expected and so I paused that for the moment. I will keep thinking about that while you review this, so let me know if you come up with other possible ways to reduce the "large SE issue" (@diegogentilepassaro I remember we discussed something along the lines of what I tried for example).
Comments on the inclusion of entry cohort X period Fes can be found in [my post](https://github.com/diegogentilepassaro/min_wage_rent/issues/43#issuecomment-698963437) in #43.


diegogentilepassaro:
@gabrieleborg before I review this please see comment [here](https://github.com/diegogentilepassaro/min_wage_rent/issues/43#issuecomment-700065101). Knowing those figures will help us understand the pros of using an unbalanced panel better!

gabrieleborg:
@diegogentilepassaro working on it!

gabrieleborg:
@diegogentilepassaro here are the numbers:

**Old (balanced) panel:**

- States: 41
- zipcodes: 1143
- ever treated zipcodes: 712
- untreated zipcodes: 431

**New (unbalanced panel:**

- States: 49
- zipcodes: 3316
- ever treated zipcodes: 2080
- untreated zipcodes: 1236

diegogentilepassaro:
Thanks @gabrieleborg! Those numbers are really surprising and it seems super weird that we don't get down our SEs as we increase the number of clusters by 20% and the number of zipcodes per state by a factor of more than 2. I will take a careful look at the code and make some experiments while reviewing. 

The main alternative idea of specification that i will explore: Given that we have many zipcodes per state, instead of including zipcode FEs we could try exploiting variation within state and across zipcodes by including state FEs. In that case we will need to control for baseline zipocode characteristics to make sure that we take into account that some zipcodes are richer, etc, and we need to allow for different zipcode time paths (as we've been doing with the zipcode specific linear or quadratic trends) but it seems that it is a good thing to try to understand what is the variation that matters!

derived/zipcode_month/code/create_final_panels.do
```diff
@@ -109,4 +118,48 @@ program create_baseline_panel
...
+		g miss`stub' =  (!missing(allmissing`stub'))
+		bys zipcode (countnonmissing year_month) : gen entry`stub' = year_month[1] if miss`stub'==1
+		format entry`stub' %tm
+		drop allmissing`stub' obsno countnonmissing
+		local allmissing_tot `"`allmissing_tot' miss`stub'"'
+	}
+	gegen miss_sum = rowtotal(`allmissing_tot')
+	drop if miss_sum==0
+	drop miss_sum `allmissing_tot'
+	gsort zipcode year_month
```
diegogentilepassaro, line 154: I thought that the unbalanced panel was just keeping all data! Why do we need those lines?

derived/zipcode_month/code/create_final_panels.do
```diff
@@ -109,4 +118,48 @@ program create_baseline_panel
...
+
+	keep if (year_month >= `=mofd(td(`start_date'))' & 					///
+			 year_month <= `=mofd(td(`end_date'))')
+	xtset zipcode year_month
+
+	merge m:1 zipcode using `instub'/zip_ready.dta, nogen assert(1 2 3) keep(1 3)
```
diegogentilepassaro, line 160: Throughout this script we use a local and and if clause in main to state that we want to add demographics, but here we do it like this. I suggest that we always add them to all versions of the panel in the way we do here and that we ditch the complicated local if structure from everywhere as it would make the code much more readable.

analysis/first_differences/code/first_differences.do
```diff
@@ -152,20 +211,25 @@ program run_dynamic_model
...
 			ytitle("Effect on ln rent per sqft") 					///
 			legend(order(1 "Full dynamic model" 3 "Distributed lags model" ///
 			5 "Effects path static model" 6 "Effects path distributed lags model") size(small))
-		graph export "../output/fd_models.png", replace
+		if "`sampl'"=="bal" {
+			graph export "../output/fd_models_`type'.png", replace
+		}
+		else if "`sampl'"=="unbal" {
+			graph export "../output/unb_fd_models_`type'.png", replace	
+		}
```
diegogentilepassaro, line 232: Let's avoid the use of if and if else unless because it clutters the code a lot. For example, we could add an optional argument called name to the function, and add the name we want for each case through a string. Does that make sense?

analysis/first_differences/code/first_differences.do
```diff
@@ -101,6 +154,12 @@ program run_dynamic_model
 		
 		keep if !missing(at)
 
+		if "`fullcenter'"=="yes" {
+			levelsof b_full if _n==`w', local(baseline)
+			replace b_full = b_full - `baseline'
+			replace se_full = 0 if _n==`w'			
+		}
+
```
diegogentilepassaro, line 162: I am not sure of what is this excerpt doing, could we please discuss?

analysis/first_differences/code/first_differences.do
```diff
@@ -8,61 +8,114 @@ program main
...
+    syntax, depvar(str) absorb(str) cluster(str) type(str)
 
 	eststo clear
-	eststo reg1: reghdfe D.`depvar' D.ln_mw,							///
+	eststo: reghdfe D.`depvar' D.ln_mw,							///
 		absorb(`absorb') 												///
 		vce(cluster `cluster') nocons
 	comment_table, trend_lin("No") trend_sq("No")
 	
+	distinct zipcode if e(sample)==1
```
diegogentilepassaro, line 117: Probably we don't need this line.

diegogentilepassaro (changes requested):
@gabrieleborg, great job! 

I am through reviewing this issue. There are a few things to discuss so I will be active in case you respond!

derived/zipcode_month/code/create_final_panels.do
gabrieleborg: The idea here is that, yes, we are keeping all data but I am "dropping" the zipcodes with all missing. I say "dropping" because I am not actually dropping them unless they are missing for all rent series. This is because if, say, zip code x has all missing rent for sfcc we don't need it. But maybe it has obs for the 2br series. So I don't want to really drop it. What I do is I flag the unbalanced samples for the different rent series (sfcc, 2br, 5multi, etc), so that when we run the regression we just run it on the sample for that series (i.e. reghdfe blablablaal if sample_sfcc==1)

derived/zipcode_month/code/create_final_panels.do
gabrieleborg: yeah I know that wasn't super nice. however I improved this part of the code in #54. There I simply create a function where you specify what you want to add (demo, econ controls etc), and it's much cleaner. You can have a look [here](https://github.com/diegogentilepassaro/min_wage_rent/blob/%2350-controls-economic-shocks/derived/zipcode_month/code/create_final_panels.do) and, if you like it, you can just leave it like this so that when we merge #54 it will be improved.

analysis/first_differences/code/first_differences.do
gabrieleborg: Fine for me!

analysis/first_differences/code/first_differences.do
gabrieleborg: We probably don't need that, I think I missed it when I was ultimating the code. It basically center the plot by shifting t-1 to zero and adjusting all coefficients and SEs.

derived/zipcode_month/code/create_final_panels.do
diegogentilepassaro: Thanks for clarifying @gabrieleborg! 

I still don't get why we need them. Those rows that have a missing sfcc but a populated 2br observation will only be used in the regression of the latter regardless of whether we do if sample_sfcc==1 purely based on what rows are populated. Right?

derived/zipcode_month/code/create_final_panels.do
diegogentilepassaro: That is very neat @gabrieleborg! Let's make sure that we push that to master soon then. I can also take a look to that PR.

analysis/first_differences/code/first_differences.do
diegogentilepassaro: Your change looks good!

derived/zipcode_month/code/create_final_panels.do
gabrieleborg: yeah in the regressions probably are redundant! but I think it's good to keep track of which zipcodes we are using with each series. So if for example we want to update some summary statistics for the unbalanced panel, we know exactly which zipcodes are we using, what's the rate of missings per zipcodes, etc? In any case we should double check that results don't change without the sample_sfcc==1

diegogentilepassaro: I see your point! Let's keep it then. One last question: do we drop some zipcode? Why would we have zipcodes without any data? If we don't drop any, then let's remove line `	drop if miss_sum==0
`

diegogentilepassaro (approved):

derived/zipcode_month/code/create_final_panels.do
gabrieleborg: yeah actually the first reason it led me to write this piece of code was that the sample was incredibly large to begin with, really slowing down my Stata SE

Issue #46, opened 2020-09-25
Sampling weights


gabrieleborg:
**Goal**
Zillow data only cover 11 percent of zipcodes, mostly from urban areas. The goal is to assess the representativeness of our sample and the robustness of our results in the light of that. We will initially try that with 2 exercises: 

1. using weights that make our sample matching some population statistics (e.g. mean values for top 100 cbsa)
2. create sampling weights using a) the number of listing in Zillow panel (available) and some zipcode level information about the rent-to-ownership ratio. 

**Steps**   

*weights to match population stats*
1. set the target population
2. create weights and run analysis

*sampling weights*
1. find Zillow listing inventory (I know I saw that somewhere, I just need to find it)
2. find rent-to-ownership ratio 
3. create weights and run analysis

gabrieleborg:
First results for the weighting exercise! @diegogentilepassaro  @santiagohermo 

I created weights to match the population statistics in the top100 cbsa. I have adopted the _entropy balancing_ procedure, a multivariate reweighting method explained in [Hainmeuller and Xu, 2013](http://web.stanford.edu/~jhain/Paper/JSS2013.pdf). In synthesis this creates weights that make the sample matching target averages selected from a target population. While developed for matching treated and control units, it can also be implemented on a single sample by providing target moments for a set of variables. The weighting procedure is done [in `analysis/first_differences/code/preclean.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/Issue-46-sampling-weights/analysis/first_differences/code/preclean.do). I have selected the following zipcode level characteristics: 
- share of rental properties
- share of black population
- median income 
- share of college graduates

As target I have used the averages we recover in [`analysis/descriptive/output/desc_stats.tex`](https://github.com/diegogentilepassaro/min_wage_rent/blob/master/analysis/descriptive/output/desc_stats.tex), column 2. 

I then simply re-run [`analysis/first_differences/code/first_differences.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/Issue-46-sampling-weights/analysis/first_differences/code/first_differences.do) using the newly computed weights. 

Results seem promising: 

- The dynamic and static models all show a slightly higher and more significant impact of MW changes on rent changes (even though we do still have the SE issue): 
<img width="652" alt="fd_models" src="https://user-images.githubusercontent.com/30894427/94304977-7ab5a000-ff3e-11ea-876c-b936fd681e77.png">
[...]

Issue #39, opened 2020-06-30
Research plan after initial draft + Feedback


santiagohermo:
**[BACKGROUND]**
We are close to finishing an initial draft in #37. After closing it, we  want start working on an improved version of the research project that includes all the feedback we've received so far

**[GOAL]**
The goal of this issue is to collect the feedback received so far and settle on an action plan going forward.

santiagohermo:
I opened up this issue to start collecting feedback and discuss what to do after the draft is done@diegogentilepassaro @gabrieleborg. I wanted a place to comment on our meeting with Jesse today, but I think it is useful to collect everything in one place.

Summary of meeting of @diegogentilepassaro and  @santiagohermo with Jesse Shapiro on 6/30.
1. To do this project correctly we need to deeply understand how the Zillow data is constructed and what it represents. For example, if we capture a small share of rentals then we would worry a lot about compositional changes around MW events.
    - To complete this it is crucial to connect with Zillow. We plan to send them the initial draft once it's done, and discuss data construction. It would also be useful to get access to their microdata.
   - While useful, Jesse gave less importance to comparing Zillow to alternative rents data.
2. The final version of the paper needs a model, as many possible channels are possible in this setting. However, we shouldn't invest too much time on it for now. We should make sure we have established reliable facts first.
3. Benchmarking the size of the effect. One useful way to start is looking at the elasticity of housing to labor demand in Diamond (2016) and building from there.

santiagohermo:
Tasks after meeting on August 25th (fyi @diegogentilepassaro @gabrieleborg)

1. Finish up data
    - Contact Zillow: To better understand data and ask for more data
    - Consider some alternative to Zillow (Trulia is also Zillow)

2. Improve first differences specification
   - Read literature
   - Try alternative models

3. Benchmark magnitude of effects

4. Model

gabrieleborg:
@santiagohermo @diegogentilepassaro 

Some updates: 
1. Finish up data: 

-  We contacted Zillow asking for: a) clarifications on the data we are using; b) additional data. Waiting for a reply

-  I looked for alternative data. Unfortunately there's no much around at the zipcode level. We could try to contact Zillow competitors, but being Zillow the largest website, we will surely get smaller samples.

2. First Diff spec: 
- Diego is trying alternative models, mostly starting from [Meer and West](http://jhr.uwpress.org/content/51/2/500.abstract). 

[...]

PR #49, opened 2020-09-27
PR for #46: sampling weights

gabrieleborg:
I am sending #46 to PR.    
I adjusted the heterogeneity results by using quartiles instead of quintiles. Results are way clearer and neat. Also, as @santiagohermo  noted down [here](https://github.com/diegogentilepassaro/min_wage_rent/issues/39#issuecomment-699129678), we leave the use of listing data to control for the composition of houses to a later date. 

For comments on the weight creation process, see [this post](https://github.com/diegogentilepassaro/min_wage_rent/issues/46#issuecomment-699097135) in issue #46 .


analysis/first_differences/output/data_file_manifest.log
```diff
@@ -2,7 +2,7 @@
 File: ../temp/fd_rent_panel.dta
```
santiagohermo, line 2: why are there two `data_file_manifest`s of the same dataset?

analysis/first_differences/output/data_file_manifest.log
```diff
@@ -13,43 +13,101 @@ Key: zipcode year_month
...
-       ln_mw |    156,600    2.079598    .1327617   1.981001   2.772589
+       ln_mw |    156,600    2.079602    .1327775   1.981001   2.772589
 -------------+---------------------------------------------------------
 ln_med_ren~t |    113,375    .0472541    .4493362  -.8722228   1.906781
 ln_medrent~c |    125,644    7.334093    .3761665   6.388561   8.794067
-nonwhit~2010 |    156,600    .4549448    .2217932   .0603325   .9963306
+nonwhit~2010 |    156,600    .4607982    .2225559   .0598227   .9962162
 trend_times2 |    156,600         451    69.27985        332        570
-med_h~at_qtl |    156,600           3    1.414218          1          5
+ wgt_cbsa100 |    156,600           1      .56017   .1560848   3.844666
```
santiagohermo, line 37: It looks like the weights go usually beyond one, and the mean is normalized to one. Is this intended? Those it matter for estimation (you are using `pw`)?

analysis/first_differences/code/preclean.do
```diff
@@ -42,21 +44,26 @@ program create_vars
 	
 	gen trend_times2 = 2*trend
 
+	ebalance `weights_vars', manualtargets(.347 .124 62774 .386)
+	rename _webal wgt_cbsa100
```
santiagohermo, line 48: Incredibly simple to generate the weights! Those values there are the target means for the `weights_vars`, right? Two comments:

- It would be useful to make a note in the code saying what file I should look at to get those values. Where are they coming from?
- Could you target the variance of variables or something like that? Just curious..

analysis/first_differences/code/preclean.do
```diff
@@ -20,8 +20,10 @@ program main
 	local het_vars "med_hhinc20105 renthouse_share2010 college_share20105 black_share2010"
 	local het_vars "`het_vars' nonwhite_share2010 work_county_share20105"
 
+	local wgtvars "renthouse_share2010 black_share2010 med_hhinc20105 college_share20105"
```
santiagohermo, line 23: Why did you choose these particular 4 variables to match?

analysis/first_differences/code/first_differences.do
```diff
@@ -58,20 +58,20 @@ program run_static_model
     syntax, depvar(str) absorb(str) cluster(str)
```
santiagohermo, line 58: The changes in this file look good to me. I noticed, however, that you are implementing the reweighting in the main `first_differences` folder. Would we like to have the results both with and without reweighing? (fyi @diegogentilepassaro)

santiagohermo (commented):
Surprisingly straightforward to implement the reweighing, @gabrieleborg. Thanks! I made some inline questions to the code.

analysis/first_differences/code/first_differences.do
diegogentilepassaro: I agree with your comment @santiagohermo! We definitely want to keep both and want both to live in separate folders.

analysis/first_differences/output/data_file_manifest.log
gabrieleborg: That's prob my bad as I was running this script manually to adjust stuff. I'll run it entirely after I finish to reply

analysis/first_differences/output/data_file_manifest.log
gabrieleborg: Yeah I noticed that, and I noticed some other researcher mentioned that in some stata blog. I believe this is because the  procedure constrains the sum of weights to match the original obs, so for any obs that weights less than 1 there will be one that weights more.

analysis/first_differences/code/preclean.do
gabrieleborg: Yeah exactly. 

1. I added a note! 
2. Unfortunately it doesn't look like. You can do it when you are balancing a treated and control group. But not when you are simply targeting some population

analysis/first_differences/code/preclean.do
gabrieleborg: I chose those as a result of a tradeoff between having too many variables to match on and the matching precision. In addition those are usually the most important demographics. I wanted to add, for example unemployment rate but then I just realized it was already the same! Also, the more variable you add, the higher certain weights will be ( at least that's what I noticed), which I don't know whether it's really a good thing.

analysis/first_differences/code/first_differences.do
gabrieleborg: I am fine with that, but before doing it let's think a second about the following: 

1. Do you think are we going to show both results? Because if we use weights in our baseline spec, I think we could simply say, when we describe how we build the data, that we create weights so to match our target population and then simply always work with weights. 
2. If we split that, it means that every time we will change something in the future we will have to update two precleans and two first_diff files right? again, it depends on whether we are going to use the weights since the very beginning or simply as a robustness check. 
3. In our other PR for the unbalanced panel, I updated everything in `first_differences`. If we duplicate this, then after merging we will have to update the weighted first diff again. So maybe it's better to merge that first, update this from master and then split it in case? 

just let me know what you think, I have no strong preferences 😊

analysis/first_differences/output/data_file_manifest.log
santiagohermo: Makes sense!

analysis/first_differences/output/data_file_manifest.log
santiagohermo: Interesting! I would have guessed, since they are called "weights", we want the weights to sum to 1. Anyway, I guess it doesn't matter as `pw` takes care of making the appropriate transformation. Thanks @gabrieleborg !

analysis/first_differences/code/preclean.do
santiagohermo: Awesome! As long as we are transparent on what we do I think we are OK. I'll mark as resolved.

analysis/first_differences/code/preclean.do
santiagohermo: That makes sense. Maybe at some point we should conduct some robustness on weight construction, even more so if we keep this specification as baseline. (fyi @diegogentilepassaro)

analysis/first_differences/code/first_differences.do
santiagohermo: Those are very good points @gabrieleborg! Maybe we can use a single `first_difference.do` and output results with and without weights. We could modify a bit the function and allow for weights to be selected from `main`.

gabrieleborg: also, something intermediate could be to have one `preclean.do`, and then `first_difference.do` and `first_difference_wgt.do`

analysis/first_differences/code/preclean.do
diegogentilepassaro: Thanks @santiagohermo and @gabrieleborg! One cool robustness to try eventually: (1) construct weights only based in demographics (share of blacks, unemployment, college share, med income) (2) construct weight only based in rental or minimum wage earners variables (share of renters, share of minimum wage earners). If we get more or less the same targeting two very different set of things seems like we are pretty robust to any weighting.

analysis/first_differences/code/first_differences.do
diegogentilepassaro: Thanks for the discussion guys! I think the wisest for now is to duplicate the preclean and have to separate folders and then open an issue to remove redundancies by (i) either construct a single preclean somewher in derived (ii) or by making sure that we have all the different panels with different variables that we need at the derived stage and in preclean files we only put derived panels together at a need basis. In this case maybe we have at derived our baseline panel, and another one that is called zipcode_demographics, and in the preclean of first differences we only touch the baseline, and in the preclean of the first differences weights we touch both. In this context, it would make sense to also split to a new folder the heterogeneity analysis, which seems a good idea anyways to have isolated as we might still need to experiment quite a bit with how to present that output. 

What we certainly don't want is to be calling things at analysis that don't come from derived and what we want to minimize is redundancy.

gabrieleborg: Ok! shall I do it or you do it? @diegogentilepassaro

diegogentilepassaro: Sorry @gabrieleborg but I had to present in CAF the trains paper and I also have my thesis deadline tonight at 12 so I am wasted, and tomorrow I am full of meetings (Applied Breakfast, Guille Alves and his colleagues from CAF, Micro Seminar, and our weekly meeting). Will only be able to get to this Friday morning at the earliest, so I guess it depends how on a hurry we are!

gabrieleborg: Hey! yeah I'll do it this morning then 👍🏽

santiagohermo (approved):

santiagohermo:
I approved the changes @gabrieleborg. Thanks!

diegogentilepassaro (approved):

Issue #58, opened 2020-10-22
Composition Bias


gabrieleborg:
**Background**
changes in rents may be caused by changes in the underlying composition of rentals in a given zipcode (entry and/or exit of different units). This might be correlated with unobserved determinants of MW changes and therefore bias our results. 

**Goal**
Address the composition bias issue both analytically and theoretically. 

**Steps**

1. Add listing inventory time-series to main dataset
2. perform "placebo" regression on such outcome
3. update Overleaf draft with better explanation of our dependent variables's strengths and theoretical motivation

gabrieleborg:
Hey @diegogentilepassaro, 

I started running some analysis replicating our models while using the (log) number of listings in the Zillow inventory as dependent variable. 

First, in [`derived/zipcode_month/code/create_final_panel.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/%2358-composition-bias/derived/zipcode_month/code/create_final_panels.do) I add to the final dataset the following variables: 

- number of listings sfcc
- number of new listings sfcc 
- list price sfcc (temporarily, to experiment)

I then created a new `analysis` folder called `first_differences_nlist`, where once again we have one [`preclean.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/%2358-composition-bias/analysis/first_differences_nlist/code/preclean.do) very similar to the others, and an [`estimate_compbias.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/%2358-composition-bias/analysis/first_differences_nlist/code/estimate_compbias.do). Here I replicate the usual analysis, by estimating all our static and dynamic models but using the log of the number of listing as dependent variable. 

Here are the main dynamic coefficients:
<img width="924" alt="fd_models_nlist" src="https://user-images.githubusercontent.com/30894427/97056768-c23c4580-1557-11eb-868c-a9d47a3d03b5.png">

the coefficients are almost never significant. However, they "jump around" a bit. Additionally, it looks like there might be some minor concerns about potential anticipation of the MW change, as shown by the increase in listings 3 months before the change. 

I also quickly run the analysis on the "unbalanced panel" (even though i haven't uploaded the scripts yet): 
<img width="924" alt="fd_models_nlist_unbal" src="https://user-images.githubusercontent.com/30894427/97056963-32e36200-1558-11eb-8b7f-6e0337706c80.png">

[...]

PR #61, opened 2020-10-26
PR for #58 composition bias

gabrieleborg:
@diegogentilepassaro sending #58 to PR after having implemented the check you suggested. 

I have tested for the impact of changes in inventory in `analysis/first_differences_nlist`. As previously explained, we have both a [`preclean.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/%2358-composition-bias/analysis/first_differences_nlist/code/preclean.do) performing the usual cleaning of the dataset, and [`estimate_compbias.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/%2358-composition-bias/analysis/first_differences_nlist/code/estimate_compbias.do) that carries out the analysis.  

I have created the following table (need to add some description too)

Screen Shot 2020-10-26 at 3 11 26 PM

- col1: our baseline model (depvar = D ln(medrent))
- col2: baseline + covar: D ln(n_listings)
- col3: placebo (depvar: D ln(n_listings))
- col4: placebo + covar: D ln(med listing price)

from the static persepctive, results don't look super nice. HOwever, when you look at the dynamic patterns, you can see that our baseline model does not really change, while the placebo one does not really detect any sudden jump at t = 0: 

**Baseline + covar**
fd_model_nlistcontrol

**Placebo & placebo + covar**
fd_placebo_comparison

As for the weighting, I gave up for the moment as it sounds just very complicated even to understand. I am sending this to PR, feel free to experiment and let's agree on a final version of tables and plots!


analysis/first_differences_nlist/code/estimate_compbias.do
```diff
@@ -0,0 +1,180 @@
...
+
+	* Static Model
+	run_static_placebo, depvar(ln_med_rent_psqft_sfcc) placebovar(ln_n_listings_sfcc) covars(ln_med_list_psqft_sfcc) ///
+						absorb(year_month i.zipcode) cluster(statefips)		
+						
+	esttab * using "`outstub'/fd_table_placebo.tex", keep(D.ln_mw) compress se replace substitute(\_ _) 	///
+			coeflabels(`estlabels_static') ///
+			stats(r2 N, fmt(%9.3f %9.0gc) ///
+			labels("R-squared" "Observations")) star(* 0.10 ** 0.05 *** 0.01) ///
+			mgroups("$\Delta y_{it}=\Delta$ ln(Median rent)" "$\Delta y_{it}=\Delta$ ln(No. listings)", pattern(1 0 1 0) ///
```
diegogentilepassaro, line 25: Let's remove the \Delta y_{it}= from the mgroups, as it is understood that it is the dependent variable.

analysis/first_differences_nlist/code/estimate_compbias.do
```diff
@@ -0,0 +1,180 @@
...
+	local outstub "../output"
+
+	use "`instub'/fd_rent_panel.dta", clear
+
+	make_results_labels
+	local estlabels_dyn "`r(estlabels_dyn)'"
+	local estlabels_static "`r(estlabels_static)'"
+
+	* Static Model
+	run_static_placebo, depvar(ln_med_rent_psqft_sfcc) placebovar(ln_n_listings_sfcc) covars(ln_med_list_psqft_sfcc) ///
```
diegogentilepassaro, line 18: Just taste: can we use controls instead of covars?

analysis/first_differences_nlist/code/estimate_compbias.do
```diff
@@ -0,0 +1,180 @@
...
+		absorb(`absorb') ///
+		vce(cluster `cluster') nocons
+
+	eststo: reghdfe D.`placebovar' D.ln_mw,	///
+		absorb(`absorb') ///
+		vce(cluster `cluster') nocons
+
+	eststo: reghdfe D.`placebovar' D.ln_mw D.(`covars'), ///
+		absorb(`absorb') ///
+		vce(cluster `cluster') nocons
```
diegogentilepassaro, line 55: Main point here: It is a little weird to have this 4 models in the same table. I think we should show just the baseline and the analogous placebo (columns 1 and 3). Column 2 is weird to show here because it is a robustness and not a placebo that should live in the table that we control for a bunch of stuff. Column 4 is confusing, first because people will expect it to be the analogous of column 2 but for a different dependent variable and that is not what it is, and second, because we are only controlling for the list price (one particular control) without controlling for many things that also correlate with both that and the number of listing (economic shocks, building permits, etc). I think that we probably just ditch column 4 from the analysis.

analysis/first_differences_nlist/code/estimate_compbias.do
```diff
@@ -0,0 +1,180 @@
...
+	store_dynamic_coeffs, model(base) w(`w')
+
+	eststo: reghdfe D.`placebovar' L(-`w'/`w').D.ln_mw, ///
+		absorb(`absorb') ///
+		vce(cluster `cluster') nocons
+	store_dynamic_coeffs, model(placebo0) w(`w')
+
+	eststo: reghdfe D.`placebovar' L(-`w'/`w').D.ln_mw D.(`covars'), ///
+		absorb(`absorb') ///
+		vce(cluster `cluster') nocons
```
diegogentilepassaro, line 74: I think that showing so much placebo is a little overkill, and the table is very clear. however, I am fine with keeping the dynamic placebo if you feel strongly about it but then, because of the reasons just stated in the comment above, let's ditch the version in which we control for the list price covariate.

diegogentilepassaro (changes requested):

diegogentilepassaro (changes requested):
Great job @gabrieleborg! Let's see if we can kill this and have the draft out there at least for the professors soon :D

analysis/first_differences_nlist/code/estimate_compbias.do
gabrieleborg: Agree

analysis/first_differences_nlist/code/estimate_compbias.do
gabrieleborg: this makes sense. For the moment I will just update the table with having those 2 columns only. When you review it tomorrow, let's think then whether it makes sense to have such a small table, or whether we can show this results somewhere else.

analysis/first_differences_nlist/code/estimate_compbias.do
gabrieleborg: I would maybe put that in the appendix? the reasoning here is that by only looking at the static table people may be mislead by the magnitude of the estimate, while if we also show the dynamics this helps in putting things back in perspective since it shows how many estimated leads and lags also have very large SEs and "jump around" a bit. what do you think?

gabrieleborg: agree in removing the plot with control.

analysis/first_differences_nlist/code/estimate_compbias.do
diegogentilepassaro: Sounds good! Keeping the placebo with 2 columns is great. Let's remember considering adding to the robustness part a column controlling for the number of listings.

analysis/first_differences_nlist/code/estimate_compbias.do
diegogentilepassaro: That sounds good @gabrieleborg! :D

diegogentilepassaro (approved):

Issue #40, opened 2020-09-04
Heterogeneity & Salience


gabrieleborg:
**Overview**
John's comment : _"If these results are correct, then I think you also need to do a bit of broader thinking on how exactly these effects can be so big.  For instance you need to think about the fraction of income that people spend on rent, do more on how the effects vary across the income distribution, and think about why the incidence would appear in this way."_

**Goal**
To better study the presence of heterogeneous effects of MW changes on zip codes. We want to understand whether different zip codes present different _transmission_ rate from MW to rent (e.g. whiter, less-educated, poorer, etc). A focus shall be put on differences in income as this determines, along with the average rent, the share of income spent on rent. 

**Steps**
- Read literature on heterogenous effects in panel data and event studies. 

- identify suitable demographic characteristics to represent zip codes. 

- test specification. 

**Note**
[This](https://economics.ucsc.edu/research/downloads/JOLE_ChatterjiChayFairlie_TheImpactofSetAsides_March18th2013.pdf) paper by Ken Chay and coauthors show a way to define heterogeneous effect in panel data sets.

gabrieleborg:
Hey @santiagohermo @diegogentilepassaro ! 

I started working on heterogeneity of the MW effect on rents. As discussed, I started working on our preferred specification, the first difference one. I explored how the effect varies across different dimensions characterizing zip codes, and I implemented 2 different approaches. 

**Demographics Variables**
Besides the demographics we already had (now correctly computed after I reviewed [base/demogrpahics/code/make_demograhics_ziplevel.R](https://github.com/diegogentilepassaro/min_wage_rent/blob/Issue40-Het-and-Salience/base/demographics/code/make_demographics_ziplevel.R)), I also computed the rent to income ratio. I believe this is going to be an important feature to look at, as it will inform us on the salience of such effect identifying those zip codes where residents pay a lot for every dollar they earn. To do that I had to combine a "fix" characteristics (as recorded in the 2010 census), income, and a time-varying one, rent. To do that I selected one month in our panel and looked at the cross-sectional distribution. Specifically, I selected January 2014 because it's the first month where we have at least 90 percent of non-missing rent observations in our final baseline rent panel. I then divided annual income by 12 and divided the monthly rent for the "monthly" income. Here is the distribution of the ratio I obtained: 
![image](https://user-images.githubusercontent.com/30894427/92792817-df3c0100-f37b-11ea-9776-b1ff29181851.png)
[...]

Issue #59, opened 2020-10-22
LODES-based salience analysis


gabrieleborg:
**Goal**
Strengthen our identification and narrative about the mechanism by showing how the effect shows up also in non-treated zipcodes where MW workers from treated zipcodes live. 

**Steps**

1. For each event, compute for each zipcode A the share of MW workers that work in the treated zipcodes but live in zipcode A. 
2. For each zipcode, compute the "experienced MW": a weighted average of the MW based on where MW workers work. 
3. Analysis: what happens when you take into account "treatment intensity" as measured by either 1 or 2

gabrieleborg:
@diegogentilepassaro @santiagohermo first pass on the LODES update. 

Since this is a rather complex technical task, I am now posting an initial description of only a part of the whole exercise. Specifically, I have now built an origin-destination matrix at the USPS zipcode level. This lives in `base`: [`base/lodes/code/make_odmatrix_ziplevel.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/%2359-LODES-based-salience-analysis/base/lodes/code/make_odmatrix_ziplevel.R). 

First of all, the LODES data contained in the OD files are not as specific as the _Workplace Area Files_, or the _Residence Area Files_. This means that we cannot observe, as before, the number of workers earning less than $1250/month **and** 29yrs or younger. In the OD files you can only observe **either** one or the other characteristic. 

The main difficulty of this exercise is that the matrix (which is actually composed by many state-level files) comes at the block level and needs to be aggregated. I adopt the following steps for each state separately (with `lapply()` in R):   

1. I collapse the count of _total jobs_, _low income jobs_, and _younger workers jobs_ at the tract level. This is done because our main crosswalk to get to USPS zipcode is a tract-zipcode crosswalk. This can only be done in a 2-step procedure: 
    - for each _origin_ block, sum target counts by _destination_ tract.
    - for each _destination_ tract, sum target counts by _origin_ block. 
2. Now that we have a tract-level od matrix, I replicate the same procedure using the tract-zipcode crosswalk with the only difference that now instead of taking the sum, I take a _weighted sum_ using the provided weights:  
    - for each _origin_ tract, weight sum target counts by _destination_ zipcode.
    - for each _destination_ zipcode, weight sum target counts by _origin_ zipcode. 

I finally save each state zip-level od matrix separately because of size: if we append them all together we would have a 34M rows dataset. 

[...]

PR #62, opened 2020-10-27
PR for #40 het and salience

gabrieleborg:
Sending #40 to PR. Please read the thread [here](https://github.com/diegogentilepassaro/min_wage_rent/issues/40) to find description of what I have done so far (essentially, all heterogeneity results so far comes from [`analysis/first_differences/code/estimate_heterogeneity.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/Issue40-Het-and-Salience/analysis/first_differences/code/estimate_heterogeneity.do)


santiagohermo:
Task received @gabrieleborg. Thanks!

diegogentilepassaro:
@santiagohermo I will take over this PR! :D

santiagohermo (commented):
I will leave this PR to @diegogentilepassaro then @gabrieleborg. I'll just leave the comment here I had made before the meeting, and mention to more things:

- As I said in the paper, I got the sense that we are compute too many tables which won't be used. I think it would be more efficient to just select the ones we want
- I made a few style changes in `make_lodes_ziplevel.R`

lib/stata/min_wage/ado/event_plot_untreated_het.ado
```diff
@@ -0,0 +1,15 @@
+program event_plot_untreated_het
```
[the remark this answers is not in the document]
diegogentilepassaro: I agree!

base/demographics/code/tract_avgMW_13_18.R
```diff
@@ -0,0 +1,77 @@
+remove(list = ls())
+source("../../../lib/R/load_packages.R")
+source("../../../lib/R/save_data.R")
+
+load_packages(c('tidyverse', 'data.table', 'matrixStats'))
+
+main <- function() {
```
diegogentilepassaro, line 7: Not sure I understand what this new script does. Why do we need the MW average from 2013 to 2018 at the tract level?

base/demographics/code/tract_avgMW_13_18.R
```diff
@@ -0,0 +1,77 @@
...
+  
+  #only problem now is that the base/demographic folder must run AFTER the base/zillow_min_wage, as it uses its output
```
diegogentilepassaro, line 9: Can't we make it part of base/zillow_min_wage? If it is based on min wage data, then it should live there.

diegogentilepassaro (changes requested):
Thanks for pushing this @gabrieleborg! I agree with @santiagohermo that we have many tables and figures that we don't use and that we won't probably use either. I propose to keep:
- Table 5 (in the current version of the draft)
- Figures with quartile-by-quartile estimates of the main 5/6 demographics (the ones in Table 5 plus a couple more at most). 
- Everything that outputs something using LODES data, so that we decide what to keep out of those in a follow-up task.

lib/stata/min_wage/ado/event_plot_untreated_het.ado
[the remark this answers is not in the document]
gabrieleborg: Agree! Done.

base/demographics/code/tract_avgMW_13_18.R
gabrieleborg: @diegogentilepassaro This was the first attempt at identifying MW workers using ACS and census data I tried. Basically, the logic was that I first compute, for each zipcode, the average MW in those years; then I take binned income data for each zipcode and assess, approximately, the share of workers earning less or equal that threshold for each zipcode. 

In the end we switched to LODES data for the draft. I think it's useful to recall the pros and cons of these two approaches, so that we can make an informed decision: 

_Census-based method_
    - PROs: census (ACS data really) provides a binned income distribution with finer categories, so in this sense it makes it easier to approximate the real share of MW workers
    - CONs: there's no information on the joint distribution income/age, which really helps finding MW workers as they mostly are low-income AND young

_LODES-based method_
    - PROs: it provides info on the share of low-income AND young (even though is only below 30 yrs old) workers for each zipcode
    - CONs: low-income can only be proxied by people earning less than $1250/month

base/demographics/code/tract_avgMW_13_18.R
gabrieleborg: If we end up keeping it, that could be a solution @diegogentilepassaro. Let me know what you think about my comment above, and then let's decide!

base/demographics/code/tract_avgMW_13_18.R
gabrieleborg: UPDATE: I re-run the heterogeneity analysis creating plots for the ACS-based MW worker variables but results are not that neat (you actually cannot spot any noticeable difference): 
fd_static_heter_sh_mww_wmean2
fd_static_heter_sh_mww_renter_wmean2

In the light of this, I think it's then better to identify MW workers with LODES data and discard this. I will wait to hear your feedback first though. One reason I think the results are that different is that I compute the ACS-based vars with a more complex procedure with many assumptions (see [`base/demographics/make_MWrenters_ziplevel.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/Issue40-Het-and-Salience/base/demographics/code/make_MWrenters_ziplevel.R)) that I believe ends up casting a "wider net": the measures are to broad to actually track MW workers?

gabrieleborg:
@diegogentilepassaro I implemented the changes requested. I ended up producing a few more het plot, but you can decide which one to keep!

base/demographics/code/tract_avgMW_13_18.R
diegogentilepassaro: @gabrieleborg, thanks for describing the pros and cons and for updating the heterogeneity analysis. A few comments/questions: 

1. Since the ACS is a very well-known dataset we still might want to keep it. We could try to fix the time dimension and instead of computing the average MW across some years, we could instead choose the initial year (2010) and compute the max MW in each zipcode for that year. Then compute the share of people falling below that income. 
2. As for this last computation of the share of people falling below some income level: the ACS has a variable called INCWAGE that comes at the annual level. We should divide it by the number of weeks that people works in a year, and the by the usual number of hours that people work in a week to obtain an hourly rate comparable with MW. Are we doing that? 
3. Are we using the proper ACS weights for the computation in 2.? 
2.  Are we sure that the ACS doesn't have age? That seems really weird!

gabrieleborg: @diegogentilepassaro here are some replies, I will work on those!

1. I can try that (if it's not too long). My only concern with that was that the bulk or our events is in the second part of the panel (2015-2019) and so I was fearing that using info from 2010 would return too imprecise measures (as opposed to using more recent data like we are doing now) !
2. RIght now we are doing the opposite, meaning bringing the minimum wage to an approximate annual total. This is because in the ACS we have data on the binned distribution of income, which I I thought it was more precise than using either an average or a median. The logic we are using is: compute, for each zipcode, the MW thresholds of annual earnings based on certain assumptions (e.g. number of MW workers in household, number of hours worked); and then get the share of residents that fall below those thresholds by looking at which bin of the income distribution is below the MW levels we find. That sounds as an alternative solution that we should definitely try, especially because what we are doing now is not very informative in the end!! the only thing is that I don't know whether I will have time to do it now, also considering that results with the LODES data are already more neat. 
3. I will double check that! 
4. We do have some actually, and I just checked that we cumputed some approximate share of MW workers based on BOTH age and income. I will review the analysis part once again to check where they end up.

diegogentilepassaro: Thanks for the replies @gabrieleborg! We can pause this and continue after #59 is sent to PR and then decide if we want to keep it. My replies:

1. Good point. Let's say we pick 2015 then.
2. We can check if the variables for number of weeks worked in a year and usual hours worked a week are nicely populated for 2015. If they do we should try!
3. 👍 
4. 👍

santiagohermo:
Any open issues here @gabrieleborg? Let me know if I should take a look at anything in particular so that we merge this back 👍

gabrieleborg:
Hey @santiagohermo ! mm not really, last time I implemented changes @diegogentilepassaro asked so I think we were good to merge. As I say in my last post, right now we are producing probably more plot than we'll use (we're producing a table with the main demo we want to discuss, but plots for more since it's easier to simply combine the selected ones in a figure) but I think it's good to have them for now so we can quickly decide what to show. 

I guess you can maybe re-run this updating the figure stile to make it ore similar to the main ones :) ! the only thing you have to be careful is starting from the right `derived` output because, as we discussed yesterday, the one produced in master does have an error in the demographic construction. I would suggest you update your drive with the latest dataset I produced in `derived_large` that I checked are correct!

gabrieleborg:
Actually there's one thing you could try @santiagohermo: I have recently updated the construction of ACS-based share of MW workers (to get a better benchmark estimate). The are variable, already in the latest `derived_large/baseline_rent_panel.dta` under the names: `sh_mww`, `sh_mww_ft`, and `sh_mww_pt`. you could try to run the heterogeneity analysis on them, because at the moment it didn't work on the older versions (see [this older post](https://github.com/diegogentilepassaro/min_wage_rent/pull/62#discussion_r515507906) here). At the moment the heterogeneity analysis works only with LODES-based MW workers definition, which is ok since we then are using that definition to build the experienced MW, but it would be interesting to see whether now we can get similar results with this new ACS-based measure of the share of MW workers living in a zipcode.

santiagohermo:
I was finally able to complete a run of `lodes/base` after updating this branch, @gabrieleborg. Changes (so that you are aware of them):

- I added a note in [LODES's readme](https://github.com/diegogentilepassaro/min_wage_rent/blob/2ef8feb6cebd39f276b7ec072ca842a8580dbd86/raw/lodes/readme.txt) about conflicting files for Alaska and South Dakota
- I moved data file manifest s to the repo (see [output folder](https://github.com/diegogentilepassaro/min_wage_rent/tree/2ef8feb6cebd39f276b7ec072ca842a8580dbd86/base/lodes/output)). Before they were being saved in the drive base_large
- I set the number of cores to 1 [here](https://github.com/diegogentilepassaro/min_wage_rent/blob/2ef8feb6cebd39f276b7ec072ca842a8580dbd86/base/lodes/code/make_lodes_experienced_ziplevel.R#L15). It looks like parallelization with `mcapply` breaks in Windows if you use more than 1 core :(
- I updated LODES and demographics files in Google Drive.

I'm now building `derived/zipcode_month`. I'll later update those files in Google Drive and re-run the relevant analysis folders. Once I'm done I'll merge. Let me know if I'm missing anything!

gabrieleborg:
Thanks @santiagohermo ! 

That Is fine, but I am just realizing now that I think all the lodes-related part of the analysis should live in #65 (via master probably), since we decided to split the lodes-based part of the analysis (revolving around the construction of experienced MW) and the heterogeneity analysis, based on census data. This is also why yesterday I was struggling to find the lodes crosswalks in #65 I guess! In this sense, I think that we should remove the heterogeneity analysis based on lodes data from here (see [this line](https://github.com/diegogentilepassaro/min_wage_rent/blob/6edaf8e5cb387d26939c09f44b1a8651e3454aad/analysis/first_differences/code/heterogeneity.do#L15) in `analysis/first_differences/code/heterogeneity.do`) and produce those plots in `first_differences_expmw/code/figures.do`. This is what I was working on, in #65. What do you think?

santiagohermo:
I think that's a great plan @gabrieleborg! I'm merging here and you can update #65 and make the changes there

Issue #64, opened 2020-10-29
Maps


gabrieleborg:
**Goal**
create some graphical visualization of the zipcodes we use, and of the variation in rent and MW.

santiagohermo:
Per meeting, here we want to:

- Make one map of the US that shows all zipcodes in Zillow data
- Find meaningful examples to plot at the MSA level. Ideas: MSAs that border across states, MSAs with some city that has higher MW than the state

gabrieleborg:
Alright guys @santiagohermo @diegogentilepassaro, sorry for the delay but it took me longer than expected to tweak all different details that go into maps..! 

First, I created a main national map showing the location of all zipcodes present in our rent panel (we have a total 3316 zipcodes): 
<img width="893" alt="Screen Shot 2020-11-08 at 1 36 49 PM" src="https://user-images.githubusercontent.com/30894427/98481319-75977200-21c7-11eb-97dd-74e79a493eb9.png">
A minor thing to remember here is that we are dropping from the pic, for clarity reason, 26 zipcodes located in Alaska and Hawaii. 

The most difficult part was to then create "zoomed" maps for specific areas. Here I created, for each area, a set of 3 maps: 
1. a map showing the percentage change in rents
2. a map showing the percentage change in actual MW 
3. a map showing the percentage change in experienced MW

Unfortunately, these map are not very informative I believe. Even if I spent 2 days trying to tweak parameters to make them clearer, I don't think we can clearly see the correlation between rent and MW increases. The main reasons I believe are more than one: first, the estimated rent increase is rather small and this means that it is therefore hard to see it on a small subset of zipodes. Second,  I spent a lot of times trying to find the right city/area to plot, but I noticed that many places where we have a local MW change (either county or city), they also have a simultaneous statewide one. So in these cases, the real variation we are using it's in the intensity of the change, but again, this is very hard to visualize on a small subset of zipcodes from a single area. 

As an example, I show you here the plots for the Seattle MSA: 

**actual MW**
![Seattle_msa_mw](https://user-images.githubusercontent.com/30894427/98481622-8943d800-21c9-11eb-9bb6-37991bfa8dc6.png)

**Experienced MW**
![Seattle_msa_expmw](https://user-images.githubusercontent.com/30894427/98481632-93fe6d00-21c9-11eb-927e-9e6bccc3ae2f.png)

**Rent Change**
[...]

Issue #67, opened 2020-11-03
Implement new main table and figure


santiagohermo:
[GOALS]
Per https://github.com/diegogentilepassaro/min_wage_rent/issues/39#issuecomment-721186499, we want to re-structure the presentation of our main results.

[STEPS]
1. Fix static table.
2. Change figure 1 to 2 panels.

diegogentilepassaro:
@santiagohermo I will add here a step 1.5:

Add paragraph stating why we think of housing supply as fixed, and show building permits placebo.

santiagohermo:
Hi @diegogentilepassaro @gabrieleborg! I completed a pass re-structuring the first differences folder and output (step 1 of the task).

- See [code](https://github.com/diegogentilepassaro/min_wage_rent/tree/5e30b9036aee1eaf55133e87c0a369fcffcc3a96/analysis/first_differences/code). I separated the production of figures from the main estimation.
- See [output](https://github.com/diegogentilepassaro/min_wage_rent/tree/5e30b9036aee1eaf55133e87c0a369fcffcc3a96/analysis/first_differences/output). We now produce `eps` format figures to include in the paper
- Following @diegogentilepassaro's approach in #60, I added the results to a [`staged_output`](https://github.com/diegogentilepassaro/min_wage_rent/tree/5e30b9036aee1eaf55133e87c0a369fcffcc3a96/paper_slides/staged_output) file [here](https://github.com/diegogentilepassaro/min_wage_rent/blob/5e30b9036aee1eaf55133e87c0a369fcffcc3a96/paper_slides/staged_output/output.pdf). One minor thing I'm not sure about is what colors to use. I'll think about it more later.

Let me know what you think!

gabrieleborg:
Thanks @santiagohermo!!! 

Two quick things: 

1. I am wondering whether it makes sense to consolidate the dynamic and cumulative effect of the dynamic table in just one output, by simply adding the estimated cumulative coefficients at the bottom of the dynamic table (I think Jesse suggested it once) ? 
2. Let's have a remainder to also update: `first_differences_wgt`, `first_differences_unbal` and `first_differences_nlist` once we decide which output we want from those part of the analysis (it can also be another issue, let's discuss it tomorrow I guess)

santiagohermo:
Thanks @gabrieleborg! On your comments:

1. I think this is a great idea. I'll implement it
2. Sounds good. I think it's a good idea to start tackling those after our call

[...]

PR #69, opened 2020-11-10
PR for #64 -  maps

gabrieleborg:
Sending the maps to PR @diegogentilepassaro  @santiagohermo. You can look at my last edits [here](https://github.com/diegogentilepassaro/min_wage_rent/issues/64#issuecomment-724950422). 

An important thing you'll notice is that the script does not follow the classic setup, main(), programs() structure, but a linear one. The reason is that, for unknown reasons, when I wrap all the code in the main function maps are not rendered. In that case we end up with either empty png files, or containing the legend only. I spent some time researching why but I couldn't find a solution.


analysis/descriptive/code/plot_maps.R
```diff
@@ -0,0 +1,551 @@
...
+                        zipzcta,
+                        zctamsa,
+                        out) {
+  
+  #removing alaska and hawaii for visual clarity (we have though 26 zipcodes there)
+  sample_zip <- df_data[!is.na(medrentpricepsqft_sfcc) & statefips!=15 & statefips!=2, first(.SD), by = zipcode][, .(zipcode)]
+  sample_zcta <- zipzcta[sample_zip, on = .(zipcode)][, .(zcta)][, value := 1]
+  setnames(sample_zcta, old = 'zcta', new = 'region')
+  
+  data(zip.map)
```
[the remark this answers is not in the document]
santiagohermo: Follow-up question: should we download those files and add them to the repo? Probably doesn't make much sense, since it's not our main focus to build maps

analysis/descriptive/code/plot_maps.R
```diff
@@ -0,0 +1,551 @@
...
+                              target_msa, 
+                              mwarea, 
+                              mwdate, 
+                              plotname, 
+                              df_data, 
+                              nmon,
+                              zctamsa, 
+                              zipzcta,
+                              zctaplace,
+                              out) {
```
santiagohermo, line 165: Just curious here. What is the logic of writing the arguments like this? I've seen this style a lot but I'm refusing to adopt it myself haha

analysis/descriptive/code/plot_maps.R
```diff
@@ -0,0 +1,551 @@
...
+    theme(legend.position = "bottom",
+          plot.title = element_text(size=180),
+          plot.subtitle = element_text(size = 140),
+          legend.title = element_text(size = 120),
+          legend.text = element_text(size = 80))
+
+  return(plot)
+}
+
+plot_mw_changes_city <- function(target_var,
```
santiagohermo, line 232: Do we need two functions, one for mw changes and other for rents? It sounds like we are only changing the variable but we still rewrite the whole structure.

analysis/descriptive/code/plot_maps.R
```diff
@@ -0,0 +1,551 @@
...
+  df_target <- df_target[Fyear_month >= as.Date(mwdate), ]
+  df_target <- df_target[df_target[, .I[1:nmonths2], zipcode]$V1]
+  df_target[, pct_rentch := (get(target_var)[.N] - get(target_var)[1])/get(target_var)[1], by = 'zipcode']
+  df_target <- df_target[, last(.SD), by = zipcode]
+  df_target <- df_target[!is.na(pct_rentch),]
+  df_target[, 'region' := zipcode]
+  df_target[, pct_rentch := round(Winsorize(pct_rentch)*100, digits = 2)] #winsorize at .05 and .95
+  
+  data(zip.map)
+  zcta_map <- sfheaders::sf_multipolygon(zip.map, x = 'long', y = 'lat', multipolygon_id = 'id', polygon_id = 'piece', keep = T)
```
santiagohermo, line 184: If feels like this zcta_map object can be created once in the bottom and then passed as an argument to each function.

analysis/descriptive/code/plot_maps.R
```diff
@@ -0,0 +1,551 @@
...
+          legend.text = element_text(size = 80))
+  
+  return(plot)
+}
+  
+df <- read.dta13(paste0(datadir, 'unbal_rent_panel.dta'))
+df <- setDT(df)
+df[, countyfips := str_pad(as.character(countyfips), 5, pad = 0)]
+df[, zipcode := str_pad(as.character(zipcode), 5, pad = 0)]
+df[, year_month := fromStataTime(year_month, '%tm')]
```
santiagohermo, line 311: cool!

analysis/descriptive/code/plot_maps.R
```diff
@@ -0,0 +1,551 @@
...
+                     nmon = 6,
+                     plotname = "Los Angeles MSA",
+                     df_data = df,
+                     mwarea = 44000,
+                     mwdate = '2019-07-01',
+                     zctamsa = zcta_msa_xwalk,
+                     zipzcta = zip_zcta_xwalk,
+                     zctaplace = zcta_place_xwalk,
+                     out = outdir)
+dev.off()
```
santiagohermo, line 374: A nice `lapply` here will probably make this shorter. Something like

```r
suffix <- c("", "_mw", "_expmw")
vars <- c("medrentpricepsqft_sfcc", "actual_mw", "exp_mw_totjob")

lapply(1:3, function (x) {
   png(filename = paste0(outdir, 'Los_Angeles_msa', suffix[i], '.png'), width = 7680, height = 7680)
   plot_mw_changes_city(target_var = vars[i], 
                     target_msa = "Los Angeles",
                     nmon = 6,
                     plotname = "Los Angeles MSA",
                     df_data = df,
                     mwarea = 44000,
                     mwdate = '2019-07-01',
                     zctamsa = zcta_msa_xwalk,
                     zipzcta = zip_zcta_xwalk,
                     zctaplace = zcta_place_xwalk,
                     out = outdir)
   dev.off()
   }
)
``` 

You could also loop across some vectors that store MSA names, and define all MSAs you want in one place. This would make it easier to change what maps you want to build.

santiagohermo (approved):
I completed a high level review of the code @gabrieleborg. I made some questions just to make sure I understand and made a few suggestions to simplify the code a bit. Once thing to flag is that this script doesn't follow our logic of having a main program and defining parameters and stuff within it. I guess that we may want to change at some point, but no need to do that now as we have more pressing issues!

Overall it looks great, thanks Gabri! It seems to me that it does what we want quite well. I'm approving, so that once you are done with the improvements we discussed today you just go ahead and merge.

Also, I updated this branch from `master` and from branch `#67`. Now you have the output tex in staged output ([this one](https://github.com/diegogentilepassaro/min_wage_rent/blob/75b8d2f7923dde9696ae1cd490b46f685ddfcfc1/paper_slides/staged_output/output.tex)). You can construct the figure that we want to add to the paper there, so that once we decide how to put it we just copy-paste, and maybe add some notes and stuff.

analysis/descriptive/code/plot_maps.R
```diff
@@ -0,0 +1,551 @@
...
+                      zctamsa, 
+                      out) {
+  census_api_key('6e4bf831630a20872606ce3949acd975473a1d87')
+  #censusvar <- load_variables(year = 2010, dataset = 'sf1') #to navigate census variables
+  census_target <- c(pop = 'P001001', urbpop = 'P002002', housing_units = 'H001001', house_urban = 'H002002')
+  df_census <- get_decennial(geography = 'zcta', year = 2010, state = NULL, geometry = F,
+                             variables = census_target, output = 'wide') %>%
+    rename('zcta' = 'GEOID') %>%
+    select(- 'NAME') %>%
+    setDT()
```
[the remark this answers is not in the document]
gabrieleborg: Yes, we are getting the few variables from the Census to plot the density! I am going to simplify this because: 

1. the total/urban distinction doesn't matter since we are plotting info only for MSA areas. 
2. the populaiton- and house-based maps look identical

analysis/descriptive/code/plot_maps.R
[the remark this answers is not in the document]
gabrieleborg: Question 1: yes. I find it the easiest and fastest way to get geographical data at the zcta level for the whole US

Question 2: I thought about it, and we could do it. But at the same time they're very fast and easy to download so for the moment I left it like that.

analysis/descriptive/code/plot_maps.R
gabrieleborg: Just visual clarity I guess 😁

analysis/descriptive/code/plot_maps.R
gabrieleborg: I started with 2 but I eventually converged on something very similar, you are right! I'll consolidate them

analysis/descriptive/code/plot_maps.R
gabrieleborg: good point

gabrieleborg:
Thanks @santiagohermo for the comments. Regarding your main comments [here](https://github.com/diegogentilepassaro/min_wage_rent/pull/69#pullrequestreview-529528572), please see my first post on why I did not apply the classic "main() function" structure. I will implement the changes you suggested and merge soon!

analysis/descriptive/code/plot_maps.R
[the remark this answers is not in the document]
santiagohermo: Sounds good, thanks! I agree that, as long as it's fast and not central to our paper, requesting the data from the internet makes sense.

santiagohermo:
Thanks @gabrieleborg! Sorry for missing your point in the opening comment. I think the linear structure is ok for now, we can revisit the pipeline after our deadline. 

I looked at the changes you implemented thus far and they look good. One quick comment about the legend. I think it should either be "6-month change" or "6 months change".

gabrieleborg:
@santiagohermo I implemented some final changes addressing the remaining issues: 

1. I managed to restructure the script and use the main() function structure. I just needed to wrap the plot in a print() function. 
2. I re-structured the city plot function that now works with a lapply() as you suggested. 
3. I removed the title and subtitle (reporting the MW change) from the figure because they didn't look good in the final latex. We can report the MSA name in the figure title and the MW date change in the figure notes. 

I am going on and merging the branch!

santiagohermo:
Great @gabrieleborg! I'll also delete the branch

Issue #70, opened 2020-11-11
Update Benchmarking Exercise


gabrieleborg:
**Goal**
Following Jesse's, Matt's and John's comments, we want to update our benchmarking exercise to provide a more reliable and truthful assessment of the magnitude of our effects. 

**Steps**
1. Based on our estimated effect, compute (and/or estimate) the total wage and rent bill at the zipcode level.
2. Compute the share of Total Wage Bill Change that get captured by the Total Rent Bill.

gabrieleborg:
@santiagohermo @diegogentilepassaro guys feel free to weigh-in/suggest variations and more detailed ideas on how to obtain meaningful estimates for this exercise.

diegogentilepassaro:
@santiagohermo and @gabrieleborg, I propose a set of steps for a benchmark. We can get it more complicated later by including some steps to use the LDOES data. Also, note that I am giving a recipe for an average measure, but we could compute this for each zipcode and have a whole distribution. 

Steps:
1. Compute for each zipcode the share of people that make the MW. Take the average across zipcodes, so that we have a measure of the share of minimum wagers on the typical zipcode. Call that share \omega. 
2. Compute a measure of per zipcode average monthyl income per individual (regardless of whether it is a minimum wager or not)  Take the average across zipocdes. Call that measure of average monthly individual income by m.
3. For zipcode-months that have a positive MW change, compute the average hourly MW change in dollars. Call that \Delta MW. Multiply \Delta MW times 40*4.35 gives a per month measure. Call that \Delta MMW. 
4. Call P the average population in a zipcode. Then we have that \Delta MMW *(omega*P) =  is the amount of extra monthly money that there is on average in a zipcode. That divide by P (so just \Delta MMW*omega) is the average increase in the in the average per capita monthly income at the zipcode. Call \Delta MMW*omega = \Delta m. Then we have that \Delta m/m is the percentage increase in average income. 
5. Compare the percentage increase in average per zipcode income with the percentage increase in average per zipcode rent that we estimated. 

Let me know your thoughts!

santiagohermo:
Thanks @diegogentilepassaro! That looks sensible. I think it would be useful to see that in a pdf as well.

Jesse's comment quoted from the pdf he sent:
[...]

Issue #65, opened 2020-10-29
Improve paper + incorporate comments


santiagohermo:
We have received several comments on the current version of the paper. In this issue, I will tackle some of those comments. Broadly, 

- I'll work on rewriting and make some other changes to improve clarity
- I'll answer to the point-by-point comments by Jesse: [00main.pdf](https://github.com/diegogentilepassaro/min_wage_rent/files/5461611/00main.pdf)

_EDIT Nov 26, 2020_: We will update the draft fully in this branch as well.

santiagohermo:
I started to address some of the issues here in branch `issue57`. After #63 is merged back to master, I will open a new branch

santiagohermo:
Hey @santiagohermo @gabrieleborg! I did a few things we talked about last week.

- I spent some time thinking of our empirical model, in particular in the Arellano Bond magic. I started [this note](https://github.com/diegogentilepassaro/min_wage_rent/blob/3826ee0b65fdd46415ced1e145f782d4149a3ea6/analysis/notes/empirical-strategy.pdf) to clearly write down the assumptions we make in each model.
- I did a lot of re-writing of the introduction, following Jesse's comments and my thoughts from the point above.
   - Paper with updated abstract and introduction: [00main.pdf](https://github.com/diegogentilepassaro/min_wage_rent/files/5470013/00main.pdf)
   - Version with answer to Jesse's comments in the introduction: [00main_JS_SH.pdf](https://github.com/diegogentilepassaro/min_wage_rent/files/5470017/00main_JS_SH.pdf)

Could you take a quick look at the updated introduction? I plan on sending this before tomorrow at noon. Thanks!

During the week I'll continue to address the comments we received and improving the draft.

diegogentilepassaro:
@santiagohermo please find my comments below:

- Comments on the empirical strategy notes [here](https://github.com/diegogentilepassaro/min_wage_rent/files/5470122/empirical-strategy_DGP.pdf).

- Comments on the updated version [here](https://github.com/diegogentilepassaro/min_wage_rent/files/5470123/00main_DGP.pdf). The only other comments about this is that the last paragraph of page 3 in the old version (now removed) used to convey the message that using month-to-month variation is much cleaner than using year-to-year. Can we add something like that to the new version?

santiagohermo:
Thanks @diegogentilepassaro! I implemented most of your changes. On your comments:

[...]

PR #72, opened 2020-11-26
PR for #70 update benchmarking exercise

gabrieleborg:
As discussed today, I am sending this to PR for review.


gabrieleborg:
@diegogentilepassaro I went over a quick review of the whole exercise, in the light of a conceptual error I think I was committing in the time aggregation of the datasets. I noticed this yesterday while writing the appendix part for controls (plots in #75 ). The reason it applies to benchmark too is that, in order to compute the average impact of MW on wages, We use QCEW county-quarter wage data. 

I think it's way easier to explain the issue in a call as it is rather technical. Maybe we can chat later/tomorrow. in the light of such revision, I computed the pass-through table again and I think we are getting closer to the truth in the sense that results are way more reasonable now: 

Screen Shot 2020-12-06 at 11 46 07 AM

As you can see now, the regression-based computation returns  a pass-through of approximately 20 percent. This is similar to what we find using Dube's estimates. I also briefly revised the formula to compute the first column, and it now returns a slightly lower share too (48 percent). Again, maybe it's better to discuss what I did live in a call, or after you read the code. my general take is that now I think all computations are more robust and I am more confident in the results we get. 

I now this is sort of last minute, but I think it's important to have a robust discussion of the possible pass-thorugh :)

diegogentilepassaro:
@santiagohermo FYI: 

@gabrieleborg and I just met about this PR. We have decided to re-organize the benchmark table a bit and to probably ditch the formula approach for now. 

So that you are aware, in the current table (the one living in [this comment](https://github.com/diegogentilepassaro/min_wage_rent/pull/72#issuecomment-739529849)) the first column, labeled with r, that displays the formula based numbers were not showing elasticities but instead the effects of the average changes in either MW or exp MW. That was one of the reasons for that column to be different than the others. The other reason, a more fundamental one, is that we are getting effects of the MW on the avg wages that are quite smaller than the ones we get from the regression-based approach or the one that Dube et al gets. For now, we are ditching the formula approach and we will probably revisit once we have time! 

Given that the table was showing a lot of information we decided to refactor and do it in the following way (TBC means to be computed):
Diego shared a sketch with you

I will move on. to review the writing in the draft and we will wrap up this PR later today!

santiagohermo:
Thanks @diegogentilepassaro @gabrieleborg! Your decisions make sense to me. Looking forward to taking a closer look to this!

Just flagging that the following table may be easier to produce than two panels

image

diegogentilepassaro:
Thanks @santiagohermo that is indeed a good suggestion!

gabrieleborg:
@diegogentilepassaro @santiagohermo we have the new table! 
 
Here is a first pass: 
Screen Shot 2020-12-07 at 8 10 09 PM

Let me know if you want to apply some modifications.

santiagohermo:
That looks really nice @gabrieleborg! How did you do to pull that out of STATA?

One quick comment is that the paper you reference as Dube et al. (2019) shows up as Cengiz et al. (2019) in the paper, so I would modify that label to match the paper.

gabrieleborg:
By using your trick to add an empty local and using it to add latex commands @santiagohermo ! Yes I will change the label soon :)

analysis/estimates_benchmark/code/benchmark_test.do
```diff
@@ -7,29 +7,343 @@ set maxvar 32000
...
+	if "`treatvar'"=="ln_mw" {
+		eststo dube: qui reghdfe D.`depvar' D.`treatvar', ///
+				absorb(`absorb')        ///
+				vce(cluster `cluster') nocons		
+	}
+	else {
+		qui reghdfe D.`depvar' D.`treatvar', ///
+				absorb(`absorb')        ///
+				vce(cluster `cluster') nocons				
+	}
```
santiagohermo, line 100: Why do we have the if-else here? It seems that this is the same regression.

analysis/estimates_benchmark/code/benchmark_test.do
```diff
@@ -7,29 +7,343 @@ set maxvar 32000
...
+	local eventsample = "D.`treatsamp'>0 & estsample==1"
+
+	local Daffected_wages = 0.068
+	local Affected_jobs   = 0.086
+	local avg_Dmw     = 10.1
+
+	qui sum `mww_share_stub' if `eventsample'
+	local avg_sh_mww  = r(mean)
+
+	local effect_wage = ((`Daffected_wages' / `avg_Dmw')*`avg_sh_mww') * 100
```
santiagohermo, line 114: I would multiply inside the parenthesis to stress that `Daffected_wages` is what's defined in shares and changing.

```Stata
local effect_wage = ((`Daffected_wages' *100/ `avg_Dmw')*`avg_sh_mww')
```
Alternatively, you can just define `Daffected_wages ` and `Affected_jobs` in percents

analysis/estimates_benchmark/code/benchmark_test.do
```diff
@@ -7,29 +7,343 @@ set maxvar 32000
...
+
+	winsor2 ratio, replace cuts(0 95)
+
+	qui sum ratio
+	local ratiomean = r(mean) 
+	local ratiomean = round(`ratiomean', .01)
+	local xratiolabel = `ratiomean' + 0.2
+	twoway (hist ratio, color(edkblue%50)), xline(`ratiomean') text(1.5 `xratiolabel' "Average: `ratiomean'", size(small)) ylabel(, grid)
+	graph export `outstub'/`treatsamp'_totwage_ratio_dist.png, replace
+end
```
santiagohermo, line 344: Seems that this is not used. Comment out or drop

analysis/estimates_benchmark/code/benchmark_test.do
```diff
@@ -7,29 +7,343 @@ set maxvar 32000
...
+
+	preserve
+	keep countyfips 
+	duplicates drop 
+	tempfile samplecty 
+	save "`samplecty'", replace 
+	use `instub_wage'/mw_wage_panel.dta, clear 
+	merge m:1 countyfips using `samplecty', nogen assert(1 2 3) keep(3)
+	sort countyfips quarter
+	reghdfe `wagevar' d_`treatvar', absorb(quarter countyfips statefips) nocons
```
santiagohermo, line 145: The "statefips" absorb is probably not identified here

santiagohermo (approved):
I made a high level review and I think the computation makes sense. I'll merge this to #65, and we can apply any extra fixes there.

analysis/estimates_benchmark/code/benchmark_test.do
diegogentilepassaro: Both the piece of code that builds the formula approach and the one that transforms it to elasticity "units" are not used anymore because the effect of MW on average wages is much lower than the other approaches so we hypothesized that there is a mistake in our code or in our reasoning. For now, we can just comment it out, but probably we will want to add a column to the table [here](https://github.com/diegogentilepassaro/min_wage_rent/pull/72#issuecomment-740285767) computing those numbers using a simple model and some number s from the literature.

Issue #66, opened 2020-11-02
Estimate our baseline models with more aggregated data and compare it with our estimates 


diegogentilepassaro:
[Goal]
Understand clearly how to compare our result with Tidemann and Yamagishi, and what are the advantages of our paper over theirs.

[Steps]
1. Construct year-zipcode, month-county, and year-county data.
2. Estimate  our baseline model in those datasets.
3. Report, compare with our estimates, and conclude. 

[Notes]
- Not sure if we want to aggregate our zipcode-month data in all cases. For example, for the zillow variables we could just download county-month data as well. For the MW data we can just use the raw data before implementing our crosswalks that transforms it to zipcode. 
- Seems like aggregating on the time dimension is less of a problem than aggregating on the geographic dimension, although still tricky because the median of a month doesn't have a clear relation with the median of a year.

diegogentilepassaro:
@santiagohermo and @gabrieleborg I have opened this issue to start thinking about the comparison of our paper with Tidemann and Yamagishi. Please let me know your thoughts!

gabrieleborg:
Thanks @diegogentilepassaro for setting this up! 

I think a good starting point would be to compute a county-month model, and eventually add a county-year one. 

In terms of data, I downloaded already the county-level data for our preferred series, sfcc (note that in the raw data is called allHomes). In this way I think it would be easy to do both exercise: aggregate our zipcode-month data, and once we have a county-month panel, add the Zillow county-month series so we can easily test both. 

For the MW changes, I think we can also adapt our setup pretty quickly. For the state and county level MW changes there's no problem. For the city level changes, I have downloaded a place-county crosswalk (you can find it in `raw/crosswalk/place_county_xwalk.csv`) so we can compute, for each city that had MW changes, how much does that affect each county that the city itself span.

diegogentilepassaro:
As per meeting:

- We will start with estimating our model with data at the county-month. For rents, we will use Zilllow data that directly come at the county-month. For MW, we will use ur raw data to construct county-month measures. We will not use our zipcode-month panel at all as aggregating a median is complicated, and assigning MW changes based on that will have more measurement error.

santiagohermo:
I made a simple simulation to play with the median of full data vs mean of medians issue, @diegogentilepassaro @gabrieleborg . See the plot below where:

- Each point shows (mean of medians of subsamples, median of full data)
- Dotted line is the 45* line
- Dashed lines are max and min across medians of subsamples

[...]

PR #75, opened 2020-12-04
Placebo control

gabrieleborg:
Guys sending this quickly to PR. This is the code I use to produce plots for the dynamic model using controls as dependent variables. 

I apologize in advance if the structure is not perfect, but time is the priority now, and I can fix stylistic choices later :) 

there are 2 scripts: 

1. [`derived/county_quarter/code/make_mw_qcewcontrol_ctyquarter.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/placebo-control/derived/county_quarter/code/make_mw_qcewcontrol_ctyquarter.do) - here I build the database to use for the analysis. I actually build **two** datasets here. this is because I realized that employment is at the month level so it doesn't make much sense to aggregate it at the quarter level. Therefore (I know it will need to be adjusted, but for the moment it lives here) this script produces one county-month dataset to use for estimation on employment; and a county-quarter dataset to use for estimation on wages and number of establishment. 

2. [`analysis/first_differences_controls/code/figures.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/placebo-control/analysis/first_differences_controls/code/figures.do) - here I create the dynamic DiD plots showing the impact of MW changes on changes in each control. There is one plot per industry. In addition, since I changed the estimation for the effect on employment at the monthly level, now the plot show both the month-level and the quarter-level coefficients together, but I think it's easy to read as I added two sets of labels.


gabrieleborg:
Guys @santiagohermo  @diegogentilepassaro I created a small PR in case you want to have a quick look at the code. I will now go to #65 and start writing down an appendix note nonetheless, so that once we are 100 percent sure about this, we can merge it back to #65 and quickly add plots.

diegogentilepassaro:
Thanks for this @gabrieleborg! One quick question: I've noticed that when you do the regressions before plotting you control for the total employment, average wage, and establishment count variables. This decision was not obvious to me. 

On the one hand, they seem like bad controls because they may be affected by the MW changes. On the other hand, we zipcode and time FEs may not be enough controls for getting right the dynamics of employment or wages for a single industry around MW changes. What is your take on this? @santiagohermo any call on this?

gabrieleborg:
@diegogentilepassaro results don't really change with or without those extra controls, but I included them because I thought you suggested it last time! :D 

We can easily remove those. 

Also, right now I am aggregating zip code level mw to the county level via housing weights (number of housing units). however I was thinking that in this context it is not obvious why we should use those, as opposed to use population weight. This is because, at the geographical level, we are simply aggregating at the county-level so we want the prevalent MW for the county (we cannot take the average as in Jesse's time aggregation).In this sense, what do we want here is to obtain the statutory MW for the county, and this feels to me better approximated by a population weight?

diegogentilepassaro:
Thanks for the quick answer @gabrieleborg! I wanted to check on that because I am not really sure of what to do, but it is really useful and comforting that they give the same results! If you agree, let's then show the ones without controlling for the total variables. My reasoning for this choice is that is more what the readers would expect, and although is not great it should be fine given that our paper is not about the effect of MW on employment or wages (or more generally on the state of the economy).

As for your point about aggregation: it is a great thing to think about that! I believe that the ideal way to have a county-month MW panel would be to do something like the following:
- If the MW change is at the state level, then it is applied to all counties that belong to that state as there are no counties that cross state borders. 
- If the MW is at the county level, then it is applied only to that county.
- If the MW is at the city level (or any level smaller than the county) then we have two options. Either we ignore those changes (it should make us understate the effect so I don't think that people can really complain) or we can say that the MW prevailing at a given county is a population-weighted average of the MW prevailing at the city and the rest of the county. 

Does this make sense to you? I am not proposing to do this now necessarily, but just trying to think what seems the right thing to do. I am also noting that we can do this as part of #66 if you prefer as we need to construct a county-month MW panel there as well!

gabrieleborg:
Thanks @diegogentilepassaro! 

At the moment, we are applying the last approach you mentioned (the weighted average). This is simply because  in the interest of time I kept it that way. Since we have a zip code month MW panel, it's more immediate to aggregate it by taking weighted averages based on the allocation we made of zipcodes to counties (when we build our main dataset). Of course, maybe in the future it will make sense to try alternative aggregation methods. 

did you have the chance to read the paragraph I wrote in the draft in #65 (sorry for the split between here and the draft)? I think we need to decide whether it's worth it to add it to the paper as it is now. The reason I am saying this is that, after I revisited the aggregation process (and I think now it's the most correct version), the results look like this: 

**Business and Services:**

Screen Shot 2020-12-06 at 10 11 14 AM

**Finance:**
Screen Shot 2020-12-06 at 10 11 56 AM

**information:**
Screen Shot 2020-12-06 at 10 12 32 AM

Notice that I also kept the employment analysis at the month level since that QCEW variable is at that level.

derived/county_quarter/code/make_mw_qcewcontrol_ctyquarter.do
```diff
@@ -0,0 +1,147 @@
...
+	merge 1:1 zipcode year_month using ../temp/exp_mw.dta, nogen assert(1 2 3)
+
+	merge m:1 zipcode countyfips using ../temp/zip_cty_wgt.dta, nogen assert(1 2 3) keep(3)
+	gsort countyfips year_month
+	*collapse geography 
+	gcollapse (mean) actual_mw dactual_mw treated_mw ln_mw exp_mw ln_expmw [aw = zhu], by(county year_month) 
+	
+	merge m:1 countyfips year_month using `instub_wages'/ind_emp_wage_countymonth.dta, ///
+	nogen assert(1 2 3) keep(3)
+	merge m:1 countyfips year_month using `instub_wages'/tot_emp_wage_countymonth.dta, ///
```
santiagohermo, line 24: Could you tell me what this file is @gabrieleborg? I'm running base/qcew but it's not produced. 

Once I'm able to wrap up a full run and confirm we are correctly implementing [the recipe](https://jmslab.github.io/DataRecipes/recipes/datarecipe_107.pdf) I'll wrap up.

gabrieleborg:
@santiagohermo That was very weird!!! 

So, I did some investigation: I remember creating that file in benchmarking (this is because there we were already using the average wage series to compute the impact of MW on wages). In that branch I had indeed modified `base/qcew/code/make_county_month_panel.do` with few lines of code producing that. The first thing I tried was to update the `placebo-control` branch with master, since benchmarking has been merged. But then, when I went inspecting the script, those lines weren't there 🤔. I went back in the commit history and I found them here (commit a2bc655c0c0c169933cb3af55c0b25d97fcd82cf). I honestly don't know why they were deleted, maybe I forgot to save...sorry! anyway, I restored them directly in `placebo-control` branch. It takes 30 second to build that panel :)

santiagohermo:
Thanks @gabrieleborg ! Note that I merged the benchmarking to #65, not to master. That's why they may not show up in master.

I'll run everything now that you added those lines

gabrieleborg:
@santiagohermo I fixed labels here :) now quarters are plot on the middle corresponding month

PR #82, opened 2021-01-20
PR for #66: reviewing the geo master crosswalk and the legacy code for the aggregation task. 

diegogentilepassaro:
There is quite a bit of legacy code in this branch when comparing to master so let's be merciless in ditching whatever didn't make it to the latest draft. As for reviewing the crosswalk, I list the steps to build it, so that the things to review are clear:

- From [this source](https://mcdc.missouri.edu/applications/geocorr2018.html) I added the ZCTA-county-place crosswalk at raw/crosswalk/geocorr2018.csv and its documentation/readme file at raw/crosswalk/geocorr_text.txt.
- I used raw/crosswalk/geocorr_text.csv along with raw/crosswalk/zip_to_zcta_2019.xlsx (a crosswalk that we had before between USPS_ZIP and ZCTA, probably we should add to the repo a companion readme with the source as part of this task) to create a master geo crosswalk at base/geo_master. The output is a .dta file that is keyed by ZCTA-USPS_ZIP-County-Place that contains the number of housing units per ZCTA-County-Place. There are more USPS_ZIP than ZCTA's so there are cases of many USPS_ZIP matching to the same ZCTA. In the end, the file has 32681 unique ZCTA's. 

One cool thing to check: 
- Do the 32681 unique ZCTA's cover the entire US? They should, but let's take a US map by ZCTA and see how we do! THis should be easy from the experienced MW maps.


base/geo_master/code/build.do
```diff
@@ -0,0 +1,28 @@
+set more off
+clear all
+adopath + ../../../lib/stata/gslab_misc/ado
+
+program main
+    import excel "../../../raw/crosswalk/zip_to_zcta_2019.xlsx", ///
+	    sheet("ZiptoZCTA_crosswalk") firstrow allstring clear
+	rename (ZIP_CODE ZCTA PO_NAME) (usps_zip zcta usps_zip_name)
```
gabrieleborg, line 8: Shall we already adopt the name format we used in the rest of the repo? In this way, we won't have to modify too much all other scripts once they ned to be updated. We should have "zcta", and "zipcode" for USPS zip codes. Also, do we need the usps_name? It looks a bit redundant to me, and a column we will eventually discard once we process this downstream.

base/geo_master/code/build.do
```diff
@@ -0,0 +1,28 @@
...
+	drop if usps_zip == "96898" & zcta == "No ZCTA"
+	keep usps_zip zcta usps_zip_name
+	save_data "../temp/usps_master.dta", ///
+	    key(usps_zip) replace
+	
+	clear
+    import delimited "../../../raw/crosswalk/geocorr2018.csv", ///
+        varnames(1) stringcols(1 2 3 4 5)
+	drop metdiv10 mdivname10 afact
+	rename (zcta5 placefp) (zcta place)
```
gabrieleborg, line 18: Same comment as above. maybe we should keep the name convention we have already adopted so that we won't have to change too much once we update other scripts. By looking at [`derived/zipcode_month/output/data_file_manifest.log`](https://github.com/diegogentilepassaro/min_wage_rent/blob/master/derived/zipcode_month/output/data_file_manifest.log), I would use "zcta", "place_code",  "countyfips", "county", "statefips". Additionally, I would remove "zipname" and and "usps_zip_name"

gabrieleborg (changes requested):
Thanks @diegogentilepassaro for setting this up! 

I left a couple of inline comments mostly for the name convention to use. In general, I think this is a great starting point to then build our datasets on. I had a look at the missing zcta. to do that I quickly downloaded all zcta with their population from NHGIS. There are 33120 zctas. I then merged the unique list of ZCTAs we have in the geomaster you created and we obtain 439 missing. Of these, 131 are actually Puerto Rico ZCTAs, so it's fine. As for the remaining 308, many have a population of 0, so I suspect they are in remote rural areas. For those with positive population, I made a quick check by randomly selecting a few and looking online what/where they are. Many are university zipcodes, so I don't think it will be a problem if we don't have them since there will probably be mostly student housing offered by universities. I didn't have time now to create a map, but that could help us identifying the final remaining missing. For those, I am still thinking how they are not present (any ideas?)

santiagohermo:
Quick question, @diegogentilepassaro! Instead of solving conflicts when updating from master, would it be easier to create a new branch, copy over the new crosswalk, and simply remove this one?

diegogentilepassaro:
@santiagohermo, thanks for suggesting that idea. I definitely agree with your approach!

@gabrieleborg, I will implement @santiagohermo's fix and take care of your comments in one new branch. I'll delete this one afterward. Sounds good?

gabrieleborg:
Good for me @diegogentilepassaro !

santiagohermo:
I'm deleting the branch here @diegogentilepassaro

PR #83, opened 2021-01-20
#82 Create geo_master

diegogentilepassaro:
PR to merge the new geo_master branch to master as per #82. Will apply the name changes before merging.


diegogentilepassaro:
@gabrieleborg I applied the renaming you suggested and ordered things a bit more. I removed the names of the USPS zip codes but I kept the ZCTA ones. I also kept the names of all the geo levels available at the crosswalk. The reason for that is that this crosswalk should be the authoritative list of geo units forever. Therefore, going forward we merge it to whatever we need and It may be that we need the names of the matches to sanity check them once we go forward, or to understand which are the ones that fail. 

Once I get your OK I will merge :D

diegogentilepassaro:
I forgot to mention, maybe we are calling the names differently elsewhere. We should ditch that so that all names come from this crosswalk everywhere that there are names. Matches should happen through codes and names should only be helpful to eyeball in some particular cases!

gabrieleborg (approved):
Thanks @diegogentilepassaro, it all makes sense! you can go on and merge :)

Issue #80, opened 2021-01-13
Explore new controls/robustness following Leung (2020)


gabrieleborg:
**Goal**
Improve our specification by refining the choice of controls and choice of events (anticipated or not)

**Steps**
1. Read and understand what Leung (2020) does
2. Assess and locate additional data sources we may need

gabrieleborg:
Hi guys @diegogentilepassaro @santiagohermo! 

I reviewed the Leung (2020) 
- His main goal is to study the MW pass-through to prices. Specifically, he focuses on retail prices by using the Nielsen data. 
- Notably, he uses the same paper we use to get most of the MW changes, so we are using similar events (besides the last years). 
- He finds that a 10 percent increase in MW leads to a .6 percent increase in prices, and he finds that pass-through for other store types is not significant. He also finds that the pass-through elasticity in grocery stores in poor counties is higher than what you would expect from a purely labor shock cost (i.e. demand is fixed and supply shifts). 
- He then uses sufficient statistics to understand the determinants of MW pass-through elasticity. He derives different pass-through formulas by assuming different things, and comparing those with his results. 
- How firms, and workers react to MW changes? Who ends up paying? He identifies 3 channels that may lead to higher prices (that is what he documents): 
    - Cost pass-through: holding demand fixed, MW raises labor cost and this translates to higher prices. This however cannot fully explain the magnitude of the price increase. 
    - Demand-side pass-through: 2 channels: 1)increase in income lead to an outward shift in demand; 2) income shock have a direct impact on demand elasticity
    - He derives the elasticity formulas and estimate them using the data. 
 
My main conclusion is that his work is related to ours in a good way (similar data, similar approach), but we fundamentally ask different questions: he's researching the impact on the product market, we're looking at the impact on the housing market (also through price). In this sense we are advantaged because there are less channels that may affect prices: I feel it's easier for us to assume fixed supply in the housing market. 

[...]

PR #84, opened 2021-01-25
PR for #80 explore leung/add data

gabrieleborg:
Hey @diegogentilepassaro @santiagohermo 

I have finished adding the new data to our workflow. I am sending this to a quick PR so we can merge it and start re-structuring the repo. 

**Local Area Unemployment Statistics (LAUS) Data**

- I have built a python script that uses BLS API to download and format LAUS data directly. This build.py script lives in [`ROOT/raw/laus/code/`](https://github.com/diegogentilepassaro/min_wage_rent/tree/%2380-explore-leung-add-data/raw/laus/code).  LAUS data comes in county files, so in the first part of the script I create a list of seriesID for each county based on the areacode information from the BLS website (see [`ROOT/raw/laus/readme.txt/`](https://github.com/diegogentilepassaro/min_wage_rent/blob/%2380-explore-leung-add-data/raw/laus/readme.txt) for more details). After that, it sends a series of API calls to retrieve the data. Note that the BLS website has limits for the number of API queries you can do in one day, and on the number of series you can retrieve with every single query. For this reason, I had to get a BLS registration key that I have added to the script. We are able in this way to get all data in one sitting.  
- We obtain a county-month dataset for the unemployment rate: `ROOT/drive/raw_data/laus/cty_laus.csv`. LAUS provides data on the labor force, employment, and unemployment absolute levels too, so with some tweak of the script we could get those too.
- we have 78 observations with the following footnote: _"Data reflect controlling to interpolated statewide totals because the survey was not conducted."_ 
- After re-reading carefully the BLS website, I think that LAUS data are obtained applying the so-called _Handbook_ method to CES and QCEW place-of-work data, which are translated to place-of-residence level (see the [LAUS Estimation Methodology page](https://www.bls.gov/lau/laumthd.htm) for more info). 
- The script log is saved in [`ROOT/raw/laus/output`](https://github.com/diegogentilepassaro/min_wage_rent/tree/%2380-explore-leung-add-data/raw/laus/output)

**Zipcode-Year level Population Data**

- I started looking to download county-year population data as discussed, but then I realized that - since we were going to use ACS yearly data - we may directly use tract-level ACS data to obtain USPS zip code level values as we do with other demographics. Conditional on the same issue with the time dimension (we can only get yearly variation), we can get a better geographical match to Zillow data by using zip code-level values. You guys tell me what you think about this :)
- For each year in our sample. (2010-2019) I used the 5-year ACS sample ending in that year. So, for example, 2010 data come from the 2006-2010 ACS, while 2019 data come from the 2016-2019 5-year ACS. 
- I downloaded the ACS data from NHGIS in 'ROOT/drive/raw_data/census/tract/nhgis0055_csv/`. 
- I wrote a simple R script that puts together the population data from all the different ACS samples, crosswalk tracts to USPS zipcodes and then expand yearly observations to a month-level ([`ROOT/base/demographics/code/make_zippop_year.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/%2380-explore-leung-add-data/base/demographics/code/make_zippop_year.R)). The results is therefore a zipcode-month panel from Jan 2010 to Dec 2019. 
- The dataset is stored in `ROOT/drive/base_large/output`


base/demographics/code/make_zippop_year.R
```diff
@@ -0,0 +1,88 @@
...
+  table_clean <- lapply(table_list, format_tables, datadir = datadir, data_version = data_version, xwalk = xwalk)
+  
+  table_clean <- rbindlist(table_clean, use.names = TRUE)
+  setorderv(table_clean, cols = c('zipcode', 'year'))
+  
+  table_clean <- table_clean[!is.na(year)]
+  table_clean[, c('from', 'to') := list(paste0(year, '-01-01'), paste0(year, '-12-31'))]
+  table_clean[, c('from', 'to') := list(as.Date(from), as.Date(to))]
+  
+  table_clean <- table_clean[, list(zipcode, acs_pop, year_month = seq(from, to, by = "month")), by = 1:nrow(table_clean)][, nrow:= NULL]
```
santiagohermo, line 38: You expand this here right? Nice trick!

base/demographics/code/make_zippop_year.R
```diff
@@ -0,0 +1,88 @@
...
+  
+  save_data(table_clean, filename = paste0(outdir, 'acs_population_zipmonth.csv'), logfile = log_file, key = c('zipcode', 'year_month'))
```
santiagohermo, line 40: Should we break these lines for clarity?

base/demographics/output/data_file_manifest.log
```diff
@@ -125,16 +125,14 @@ Key:  zipcode
  ------------------------------------------------------------------------------------------------------ 
 
 ============================================================================ 
-File: ../temp/mw_avg1318.csv 
-MD5:  322bb8e21fdbcf946eebe598906a7807 
-Key:  tract_fips
```
santiagohermo, line 130: We are not producing this file anymore? Why is that?

raw/laus/code/build.py
```diff
@@ -0,0 +1,127 @@
...
+                        footnotes = footnotes + footnote['text'] + ','
+                if 'M01' <= period <= 'M12':
+                    #x.add_row([countyfips, year,month,unemp_rate,footnotes[0:-1]])
+                    county_list.append(countyfips)
+                    year_list.append(year)
+                    month_list.append(month)
+                    unemp_list.append(unemp_rate)
+                    footnote_list.append(footnote)
+        
+        df = pandas.DataFrame({'countyfips': county_list, 'year': year_list, 'month': month_list, 'unemp_rate': unemp_list, 'footnotes': footnote_list})
```
santiagohermo, line 88: Can we break this line?

raw/laus/code/build.py
```diff
@@ -0,0 +1,127 @@
...
+    print(message)
+    logger.info(message.replace("\n", ""))
+
+
+
+if __name__ == "__main__":
+    
+    AREACODE_PATH = os.path.dirname(os.path.dirname(os.path.realpath(__file__))) + '/docs/'
+    LOG_PATH      = os.path.dirname(os.path.dirname(os.path.realpath(__file__))) + '/output/'
+    DATA_PATH     = os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.realpath(__file__))))) + '/drive/raw_data/laus/'
```
santiagohermo, line 105: I think it's cleaner to define "LAUS_PATH" before and use `os.path.join` to join paths. I'd do something like

```python
    LAUS_PATH     = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
    AREACODE_PATH = os.path.join(LAUS_PATH, 'docs')
    LOG_PATH      = os.path.join(LAUS_PATH, 'output')
    DATA_PATH     = os.path.join(LAUS_PATH, 'drive/raw_data/laus') 
```

raw/laus/code/build.py
```diff
@@ -0,0 +1,127 @@
...
+import pandas
+import logging
+import os
+import sys
+import datetime
+
+def create_seriesID_list(instub):
+    
+    #read in area codes and build list of laus series to download
+    areacodes = pandas.read_csv(instub + 'bls_areacodes.csv')
```
santiagohermo, line 17: I think this CSV file is not in the repo

raw/laus/readme.txt
```diff
@@ -0,0 +1,24 @@
+OVERVIEW
```
santiagohermo, line 1: Very neat README file!

raw/laus/temp/cty_laus.txt
```diff
@@ -0,0 +1,244 @@
++------------+------+-------+------------+-----------+
```
santiagohermo, line 1: I don't think we want to commit this file. Maybe we can:

- Delete the file
- Adjust gitignore (I thought that the `temp` folder was always ignored)

santiagohermo (changes requested):
Thanks @gabrieleborg! Really nice work. 

In the inline comments I make a few questions and request a few style changes only. General questions are below:

- Regarding the `laus` data.
 - I don't see any `docs` folder, is it possible that a csv file inside there was ignore by Github?

- Regarding the population dataset:
  - it varies is keyed at the zipcode-month, but the data actually varies at the zipcode-year, right?
  - does it include all ZIP codes in the US?
 
- Finally, I noticed some changes in `raw/crosswalk/TRACT_ZIP_122019.xls`. That surprised me, did you update this crosswalk?

gabrieleborg:
Thanks @santiagohermo ! 

- Regarding the `laus/docs/bls_areacodes.csv`, I added an excpetion to gitignore, thanks for spotting that!
- as for the population data: Yes, the data varies at the year level as they come from ACS. However I expanded it already to match our preferred unit-time pair, i.e. zip-month. In terms of zipcodes, yes it should include all of them. The reason is that they originally are tract-level data from the ACS, so by default they cover all US. Then I used a tract-zipcode crosswalk to convert them to USPS zip codes so it should cover the whole US. and yes, I added a crosswalk but I didn't want to get into details there because it's a bit complicated and I wanted to discuss that in person today. Long story short: USPS zipcodes are constantly updated by the Postal Service, so they release new crosswalk every quarter. As of now, we only used one (the earliest one HUD published, i.e. Q4 2012), but I noticed that we were getting less zipcodes than when using instead a crosswalk from the last period in our sample (Q4 2019). So that's why I downloaded it. In general, we should think about this because we never really addressed the issue. 

I will nor read and address the inline comments :)

base/demographics/code/make_zippop_year.R
gabrieleborg: yes, you simply report the list of variables you want to keep and create the new date combining the from and to columns with the sequence function. And you apply this to every row with by :)

base/demographics/output/data_file_manifest.log
gabrieleborg: That's weird because I have it in my temp folder...

raw/laus/code/build.py
gabrieleborg: Added

raw/laus/temp/cty_laus.txt
gabrieleborg: They are, strange! anyway, that was just a try we don't need. I deleted it.

raw/laus/code/build.py
gabrieleborg: I updated paths following your advice

base/demographics/output/data_file_manifest.log
santiagohermo: It looks like the file was produced and the manifest moved around! (see [here](https://github.com/diegogentilepassaro/min_wage_rent/blob/64ef577e829d620b21c3909b1a8e48e7d6c7ab3e/base/demographics/output/data_file_manifest.log)) Marking as resolved :)

raw/laus/temp/cty_laus.txt
santiagohermo: Got it!

santiagohermo (approved):
Thanks @gabrieleborg! I just made a change to the crosswalk to match our approach in `raw/crosswalk` (i.e., having a folder specific gitignore file).

Go ahead with the merge whenever

santiagohermo:
I'll squash and merge here @gabrieleborg

Issue #86, opened 2021-01-26
Separate zillow and mw construction


santiagohermo:
Currently, `base/zillow_min_wage` constructs a panel with rents and MW data containing ZIP codes available in Zillow. The goal of this task is to separate these steps, and make MW panels for the entire population of US ZIP codes and US counties. We also want to compute MW measures including and excluding local MW changes (to compare with the literature)

This issue is part of the pipeline updating in #85. Branch `#86-pipeline_zillow_min_wage` was created to track changes, which will later be merged back to `#85-pipeline`.

santiagohermo:
Hey @gabrieleborg @diegogentilepassaro. I updated `zillow`. I changed the approach a bit and now we produces two "final" datasets, which are stored in `drive/base_large/zillow`

- `zillow_county_clean.csv`: stores zillow data at the county-month level
- `zillow_zipcode_clean.csv`: stores zillow data at the zipcode-month level

See `data_file_manifest.log` [here](https://github.com/diegogentilepassaro/min_wage_rent/commit/5b1ec90116c7c3b677f33847b6726b08dd3b571a#diff-1682f87071a09431b8f8fdc3d60f54b1ff84649d3b1263c0f0647249e88b78e9). Note that I did not add the datasets to the Google Drive yet (they live in my local drive).

The code is now structured in two `.R` files, [`build_zipcode.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/5b1ec90116c7c3b677f33847b6726b08dd3b571a/base/zillow/code/build_zipcode.R) and [`build_county.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/5b1ec90116c7c3b677f33847b6726b08dd3b571a/base/zillow/code/build_county.R#L20), which I built based on previous scripts `base_large/zillow/code/RenameZillowVars_zipLevel.R` and `derived/zipcode_month/code/ReshapeMergeZillow_zip.R` developed by @gabrieleborg before.

[...]

Issue #87, opened 2021-02-02
Apply Geo Master Crosswalk in Base


gabrieleborg:
**Goal**
Update `base`'s scripts so to uniformly use the new `geo_master` dataset. 

This issue is part of the pipeline updating #85. Once done, merge back to branch `#85-pipeline`.

_UPDATE_: See plan for updated pipeline [here](https://github.com/diegogentilepassaro/min_wage_rent/issues/85#issuecomment-774740656). In this issue the main goal is to update `lodes` to output origin-destination matrices. Splitting MW and Zillow stays confined to #86.

gabrieleborg:
Hi @santiagohermo and @diegogentilepassaro! 

I am posting here the initial updates to`base` where we start applying the new geo_master xwalk. As I told you, I will come back to work on this tuesday night/wednesday morning (Feb 10th), so I will try to explain in as much details as I can what I did so far, in case you want to pick up something. 

So far I have split the folder `base/zillow_min_wage` in two: `base/zillow` and `base/min_wage`. 

**`base/zillow`** 
- here we now have only one file, [`base/zillow/code/RenameZillowVars_ziplevel.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/apply-geomaster-base/base/zillow/code/RenameZillowVars_zipLevel.R). This file didn't need any update since it doesn't use any crosswalk, but it simply format Zillow data that are already at the zip code level. 

**`base/min_wage`**
- Here I moved 4 files: 
    - [`substate_mw.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/apply-geomaster-base/base/min_wage/code/substate_mw.do): nothing to update here since it simply creates substate level mw datasets (one county and one local)
    - [`state_mw`](https://github.com/diegogentilepassaro/min_wage_rent/blob/apply-geomaster-base/base/min_wage/code/state_mw.do): same here
    - [`cleanGeoRelationshipFiles.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/apply-geomaster-base/base/min_wage/code/cleanGeoRelationshipFIles.R): we are not using this script anymore since all the crosswalk between geos are already addressed in `geo_master`. However, I decided to leave it here _temporarily_. This is becasue I didn't finish to update other `base` folders (e.g. `bps`, `demographics`) and I don't remember whether there are dependencies with output of such script. Once we update everything and we are sure, we can delete it. Same logic for dependencies in `derived`. 
[...]

Issue #85, opened 2021-01-26
Update pipeline


santiagohermo:
In this issue we will track the updating of the pipeline. The main goal is to increase flexibility, so that it gets easier to construct datasets at different levels of aggregation for analysis. Some of those changes were discussed [here](https://github.com/diegogentilepassaro/min_wage_rent/issues/79#issuecomment-758906949).

Below is a running list with tasks to accomplish to complete this issue.

Sub-tasks:
| Issue | Description                                                  | Status |
|  ---    | ---                                                                | ---    |
|          | Split Zillow and MW construction in `base` |          |
|          | Apply `geo_master` crosswalk in `base`       |          |
|          | Update zipcode-month panel                     |          |
|          | Update county-quarter panel                      |          |

Notes:
- All sub-tasks will branch off of `issue85-pipeline`. 
- Each issue will be peer-reviewed and later merge back to `issue85-pipeline`. Later, this branch will be reviewed to make sure everything adds up together nicely.

santiagohermo:
I started this task and assigned all of us @diegogentilepassaro @gabrieleborg. I also opened #86 for myself. Feel free to start other tasks and modify the opening comment to keep it updated.

diegogentilepassaro:
Thanks @santiagohermo! I will assign myself updating the county-quarter panel. 👯

santiagohermo:
I want to get rid of the `load_packages` function. The times it had to install a package in batch mode it failed, so I had to install it manually anyway. What do you say if we move to a config file that installs all packages used in the repo? @gabrieleborg @diegogentilepassaro

santiagohermo:
Here is a plan for the updated pipeline @diegogentilepassaro @gabrieleborg.

- `base/min_wage` produces datasets with the mw at state and substate levels, and different time aggregations
  - [implemented in #86)
- `base/zillow` produces datasets of zillow data at the zip code and county levels
  - [implemented in #86]
- `base/lodes` produces the origin-destination matrix needed to compute the experienced MW, plues RAC and WAC statistics
  - [to be implemented in #87]
- `derived/min_wage` produces datasets of the statutory and experienced mw for the whole US at different time aggregation levels
  - uses output of `base/min_wage`, `base/geo_master`, and origin-destination matrices from `base/lodes`
- `derived/zipcode` and `derived/county` produce datasets restricted to zillow sample of rents and MW (and others) at the monthly, quarterly and yearly levels
[...]

PR #90, opened 2021-02-08
PR for #86: Separate zillow and mw construction

santiagohermo:
Hi @diegogentilepassaro! Can you review this PR for #86? You should focus on reviewing

- `base/zillow`, which now outputs two datasets at the zip code and county levels (with monthly frequency)
- `base/min_wage`, which outputs files at the state and substate levels with MW vars.

I also dropped some files in `derived/min_wage` which will be needed in the future (once we start with this task)

I'm also asking @gabrieleborg to explain to me why we need [this function](https://github.com/diegogentilepassaro/min_wage_rent/blob/5b1ec90116c7c3b677f33847b6726b08dd3b571a/base/zillow/code/build_zipcode.R#L123-L163) in `base/zillow/code/build_zipcode.R`. Why can't we just merge all datatables together @gabrieleborg, as I did with the [county data](https://github.com/diegogentilepassaro/min_wage_rent/blob/5b1ec90116c7c3b677f33847b6726b08dd3b571a/base/zillow/code/build_county.R#L20)?

Thanks!


base/min_wage/code/state_mw.do
```diff
@@ -3,29 +3,29 @@ clear all
...
-    local xwalk     "../../../raw/crosswalk"
-    local exports   "../output"
-    local temp      "../temp"
+    local raw     "../../../drive/raw_data/min_wage"
+    local xwalk   "../../../raw/crosswalk"
+    local outstub "../output"
+    local temp    "../temp"
 
-    import_crosswalk, instub(`xwalk') outstub(`temp')
+    import_statenames, instub(`xwalk') outstub(`temp')
```
diegogentilepassaro, line 11: Unimportant: there is a cool STATA package called statastates that would be useful here 🍡

base/min_wage/code/state_mw.do
```diff
@@ -3,29 +3,29 @@ clear all
...
 
     prepare_finaldata, begindate(01may1974) finaldate(31dec2019)           ///
                        outstub(`temp') temp(`temp')
 
     local mw_list "fed_mw mw"
-    export_state_daily,     instub(`temp') outstub(`exports') target_mw(`mw_list')
-    export_state_monthly,   instub(`temp') outstub(`exports') target_mw(`mw_list')
-    export_state_quarterly, instub(`temp') outstub(`exports') target_mw(`mw_list')
-    export_state_annually,  instub(`temp') outstub(`exports') target_mw(`mw_list')
+    export_state_daily,     instub(`temp') outstub(`outstub') target_mw(`mw_list')
```
diegogentilepassaro, line 22: Do we really need the daily?

base/min_wage/code/state_mw.do
```diff
@@ -3,29 +3,29 @@ clear all
...
+    export_state_monthly,   instub(`temp') outstub(`outstub') target_mw(`mw_list')
+    export_state_quarterly, instub(`temp') outstub(`outstub') target_mw(`mw_list')
+    export_state_annually,  instub(`temp') outstub(`outstub') target_mw(`mw_list')
```
diegogentilepassaro, line 25: Maybe rename to export_state_max_quarterly and export_state_max_yearly

base/min_wage/code/substate_mw.do
```diff
@@ -69,8 +69,6 @@ program substate_min_wage_change
 	label var mw "Minimum Wage"
 	order statefips statename stateabb locality year month day date mw mw_* source source_2 source_notes
```
diegogentilepassaro, line 70: Add a break line

diegogentilepassaro (approved):
@santiagohermo I just completed a review of this task. Mostly it looks great, but very soon when we apply the geo_master to base in #87 substantial chunks of code will change, as both `base\zillow` and `base\min_wage' will be directly merged to the geo_master. I suggest that we both review thoroughly there once is through.

base/min_wage/code/substate_mw.do
santiagohermo: Done

base/min_wage/code/state_mw.do
```diff
@@ -146,78 +144,75 @@ program prepare_finaldata
...
-             by(statefips statename stateabb monthly_date)
+    collapse (max) `target_mw', by(statefips statename stateabb monthly_date)
 
     label var monthly_date "Monthly Date"
     label_mw_vars, time_level("Monthly")
 
-    isid statefips monthly_date, sort
-    export delim using `outstub'/VZ_state_monthly.csv, replace
+    save_data using `outstub'/state_monthly.csv, key(statefips monthly_date) ///
+        outsheet replace
```
santiagohermo: The "using" in L176 is a typo and was breaking this script. I dropped it and now it seems all datasets build correctly

base/min_wage/code/state_mw.do
santiagohermo: Do we want to always use the max, even in the yearly version? 

(Changed "annually" for "yearly")

base/min_wage/code/state_mw.do
santiagohermo: At first I thought we don't and I removed this, but the code broke. So, it seems that we do need this. The "state_daily" is used to build the "substate_daily". It's important to build at this dimension I think cause some MW changes are not on the 1st

base/min_wage/code/state_mw.do
santiagohermo: Looks interesting, do you want to implement it in the code? :)

santiagohermo (commented):
I replied and applied some changes @diegogentilepassaro. I will wait to merge though, since I want to hear @gabrieleborg's reply to the question I made in the [opening comment](https://github.com/diegogentilepassaro/min_wage_rent/pull/90#issue-569784511)

gabrieleborg:
Hi @santiagohermo and @diegogentilepassaro! 

Sorry for the delay!

To answer your question about [this function](https://github.com/diegogentilepassaro/min_wage_rent/blob/5b1ec90116c7c3b677f33847b6726b08dd3b571a/base/zillow/code/build_zipcode.R#L123-L163): 
- The original Zillow raw series come separated, i.e. one csv for each house type. Within each file, every zip code is then identified in heterogeneous ways. For example, in one series the zipcode 501 only shows state abb, while in for another series it can show also city, or msa. 
- since we weren't using a master geofile, those lines create a uniform set of geovariables for each zip code. Example: 
Screen Shot 2021-02-10 at 10 42 51 AM
the 501 zip code has the city and msa shown only in certain series but not in others. 
- Of course, we could have simply merged all files by zip code only, but since we needed the other geo to then add min wage, I wrote those lines so to look at all zipcode, city, msa, and county combinations (for each zipcode) and keep the most populated one. I did so by keeping the one with the least number of missing infos. 
- since now we have a master geo file, I don't think we need that anymore :)

base/min_wage/code/state_mw.do
diegogentilepassaro: I would think that is the most natural approach, but we can leave a couple of options and explore if you want. What do you have in mind? This would impact the way we compare with Yamagishi and Tidemann and my intuition is that taking the max will make it such that a year is treated when in fact is only partially treated, so that they understate the effect. What do you think?

base/min_wage/code/state_mw.do
diegogentilepassaro: I gave it a shot but it is actually complicated to change our code because it iterates through states to build the state panel. I left it as is!

santiagohermo:
Thanks @gabrieleborg! [That](https://github.com/diegogentilepassaro/min_wage_rent/pull/90#issuecomment-776817056) makes a lot of sense. Then, I will simplify those lines merging by zipcode only, and then I will add the other geovars from the crosswalk.

diegogentilepassaro:
@santiagohermo I spotted a type in line 202 in `base\min_wahe\state_mw.do`. I fixed it and made a fresh run. After we finish discussing [here](https://github.com/diegogentilepassaro/min_wage_rent/pull/90#discussion_r573864500) we are ready to close.

santiagohermo:
Thanks for spotting that typo @diegogentilepassaro! I'll also quickly update zillow follow [this](https://github.com/diegogentilepassaro/min_wage_rent/pull/90#issuecomment-776845562)

base/min_wage/code/state_mw.do
santiagohermo: Since our goal with those data is to compare to Yamagishi and Tidemann, I think the approach makes sense. (I was thinking an alternative where the unit is not treated if the MW is "late" or something, but no need to spend time on that)

base/min_wage/code/state_mw.do
santiagohermo: Another typo spotted [here](https://github.com/diegogentilepassaro/min_wage_rent/pull/90#issuecomment-776846202)

santiagohermo:
I wrapped up the `zillow` folder. I basically keep only the zipcode or county from the raw data, and then use the `geo_master` to add other geographies to the final dataset.

I also added some "README" files to track depedencies. The repo structure is complicating a bit, and we may forget what to run first in the future! (We could also have some "run_base", "run_derived", .. etc to state what order we need to run things in)

I'll proceed to merge to the branch for #85.

santiagohermo:
FYI @gabrieleborg @diegogentilepassaro

Issue #81, opened 2021-01-15
First attempt on a serious model


santiagohermo:
In this issue I will start thinking about a potential model for the paper. 

Notes:
- See Jesse's guidelines in https://github.com/diegogentilepassaro/min_wage_rent/issues/79#issuecomment-756338280
- Old toy model available in `analysis/notes`

santiagohermo:
Paper to cite: [Manning and Petrongolo (2017)](https://www.aeaweb.org/articles?id=10.1257/aer.20131026). It shows that labor markets are quite local, making it important to focus on small geo units

santiagohermo:
I'll use this comment to collect what seem relevant papers.

### Urban Models

- [Ahlfeldt et al (2015). The Economics of Density: Evidence From the Berlin Wall](https://www.princeton.edu/~reddings/pubpapers/Berlin-Ecta-10876.pdf). Model of residence-workplace choice within a city that is located in some wider economy. The within-city choice is defined at the block level, utility is exogenously given by the country, and firms choose location as well. There is a housing sector that determines residence vs office space. 
- [Monte, Redding and Rossi-Hansberg (2018). Commuting, Migration, and Local Employment Elasticities](https://www.aeaweb.org/articles?id=10.1257/aer.20151507). Explores the determinants of _local employment elasticities_. Writes down model of residence-workplace choice that includes a goods market ("We develop a spatial general equilibrium model in which locations are linked in goods markets through trade and in factor markets through migration and commuting".)
- [Jorge Perez (2020). City Minimum Wages](https://jorgeperezperez.com/files/Jorge_Perez_JMP.pdf). Builds an urban model with 3 types of workers who get to choose where they live and work. There is a representative firm in each location endowed with a CES production function. There is also a matching technology for each location (so unemployment), and a housing market.

### Min wage Effects

- [Bhaskar, Manning and To (2002). Oligopsony and Monopsonistic Competition in Labor Markets](https://www.aeaweb.org/articles?id=10.1257/0895330027300). Argues that oligopsony can explain low disemployment effects.
- [Aaronson and French (2007). Product Market Evidence on the Employment Effects of the Minimum Wage](https://www.jstor.org/stable/10.1086/508734?seq=10#metadata_info_tab_contents). Early example of theoretical discussion of employment effects for different industrial structures (and empirical estimation)
[...]

PR #91, opened 2021-02-15
PR for #81: Model

santiagohermo:
Hey guys! Can you review the [model proposal](https://github.com/diegogentilepassaro/min_wage_rent/blob/253978213910c043cf1dc6124d95817156e1c679/issue81/model_proposal.pdf) built in #81? The code lives [here](https://github.com/diegogentilepassaro/min_wage_rent/blob/253978213910c043cf1dc6124d95817156e1c679/issue81/model_proposal.Rmd). I added a few things since the last time we talked. I think they need polishing, but illustrate other things we could do with this measure and how we could "sell" the paper.

Sorry for the delay! It's been a busy week. I think the goal of this PR is to polish this note and meet with professors to get their opinions on what to do with the paper. Diego and I have a presentation in the Applied Micro Lunch in two weeks, so maybe we can present it there as well to get some feedback.

Just for reference, 

- [here](https://github.com/diegogentilepassaro/min_wage_rent/blob/253978213910c043cf1dc6124d95817156e1c679/analysis/notes/model.Rmd) is the code for the more-complicated model I was working on earlier (pdf [here](https://github.com/diegogentilepassaro/min_wage_rent/blob/253978213910c043cf1dc6124d95817156e1c679/analysis/notes/model.pdf)), and 
- [here](https://www.overleaf.com/project/6011e2221cd9783d12536d10) is Diego's attempt.


issue81/model_proposal.Rmd
```diff
@@ -0,0 +1,245 @@
...
+\frac{d y_{ni}}{d \underline{w}^L} > 0 , 
+\quad 
+\frac{d y_{ni}}{d \underline{w}^R} < 0
+\quad \quad  \forall (n,i) ,
+$$
+
+so that workplace minimum wages increase disposable income of these workers, whereas 
+residence ones lower it.
+
+We denote housing demand by the $(c,i)$ group as $H_{ni} \left(r_n, y_{ni} \right)$,
```
diegogentilepassaro, line 65: $(c,i)$ should be $(n,i)$

diegogentilepassaro (approved):
Just found 1 typo! 

After the presentation, we can try a model in which we fix exogenously residence locations but with a transport cost, we allow for workplace location choice. That should be representative of the short term and should be enough to generate that the households from the same type chose different workplaces. We can have a firm that produces local good using both MW and non-MW workers and housing supply is fixed. 

That simple framework should make it easy to get welfare under fixed residence and fixed house supply. That sounds to me like a good way to start thinking about microfounding the model, instead of the old approach I tried where estimation was needed.

issue81/model_proposal.Rmd
santiagohermo: Fixed!

santiagohermo:
Thanks @diegogentilepassaro! [That](https://github.com/diegogentilepassaro/min_wage_rent/pull/91#pullrequestreview-594659002) idea sounds very interesting.

Another doubt I have is whether we should call this argument:

https://github.com/diegogentilepassaro/min_wage_rent/blob/defd77d780c7aba67d18e0782f37a27be54a8b04/issue81/model_proposal.Rmd#L185-L190

an "identification" argument. I think for that we have to say that, in the version with time, you can't include both MW measures since they are collinear or something. Just something to think about.

I'll delete this branch then, and we can go back to the [code](https://github.com/diegogentilepassaro/min_wage_rent/blob/defd77d780c7aba67d18e0782f37a27be54a8b04/issue81/model_proposal.Rmd) or [pdf](https://github.com/diegogentilepassaro/min_wage_rent/blob/defd77d780c7aba67d18e0782f37a27be54a8b04/issue81/model_proposal.pdf) in the future when writing this down somewhere.

Issue #92, opened 2021-02-16
Build MW variables in `derived/min_wage`


santiagohermo:
In this issue we will update the step `derived/min_wage` in the pipeline. This step will use the output from
- `base/geo_master`,
- `base/min_wage`, and
- `base/lodes`

to construct a series of MW measures for the entire US. In particular, we want to output two datasets (one at the zipcode and other at the county level, both with monthly frequency) that contain the following MW vars:

- actual mw: maximum between city, county, and state levels,
- actual mw excluding city ordinances,
- experienced MW, i.e., `sum_{i \in Z} \pi_{ni} MW_{i}` for each zipcode `n`, where `pi` are lodes weights and `MW_{i}` is the actual mw,
- log of above variables,
- experienced log MW, i.e., `sum_{i \in Z} \pi_{ni} log MW_{i}`

This task is tracked in #85. As such, it will branch off of `issue85-pipeline` and later merge back there.

santiagohermo:
I envision the structure as follows:

- `build_statutory.R/do` constructs MW measures for both geographies as max of city, county and state levels
- `build_experienced.R/do` constructs experienced MW from output of step above, and creates a new dataset at the zipcode level with experienced MW measures.

Happy to discuss in the call tomorrow!

santiagohermo:
One comment here @diegogentilepassaro @gabrieleborg. I noticed while taking a look at the code in #93 that, to build the exp mw, we do the following:

- compute exp_mw by state
- paste all states

Does this take into account commuters to other states? I guess not? Could we account for them in the code by first putting the entire matrix together or something?

gabrieleborg:
Hey @santiagohermo !

The script already takes care of out-of-state workers. This is done by the `aux` files that are merged together and added to each state matrix ;)

santiagohermo:
Thanks for the quick reply,@gabrieleborg! I hadn't noticed those files, that's great.

gabrieleborg:
thanks @santiagohermo for the review, it looks good to me!

yes, let's look at the experienced MW in the call :)

santiagohermo:
Wait for geo_master to have unique zip code-zcta combination. Build experienced.

Other things to do:
- Fix name of county dataset
- Keep whole US

santiagohermo:
Note: branch updated from `issue85` after fixing geo_master in #96

santiagohermo:
Continues in #100

PR #94, opened 2021-02-16
PR for #87: Update `base`

santiagohermo:
Let's review this for #87 @diegogentilepassaro @gabrieleborg. 

I think the main things is to run the steps in `base`, make sure they work and produce the datasets we want, and confirm that the Google Drive is up to date with the changes. Doing a review of the changes in the code will also be useful, although bear in mind these are mostly changes on stuff we already had reviewed before.


diegogentilepassaro:
Task received @santiagohermo!

base/safmr/code/build_zipcode.do
```diff
@@ -0,0 +1,240 @@
...
+	prepare_data, instub(`raw')
+
+	forval fmr = 2018(-1)2012 {
+		append using ../temp/safmrs_`fmr'.dta
+	}
+
+	drop if missing(zipcode)
+	sort zipcode year
+
+	drop if zipcode < 1000
```
diegogentilepassaro, line 19: Are we sure this is OK? The ideal would be to have all zipcodes as strings everywhere in the repo, but even if they are integers what happens to a zipcode like [this one](http://www.maphill.com/united-states/alaska/00001/location-maps/physical-map/)?

base/safmr/code/build_zipcode.do
```diff
@@ -0,0 +1,240 @@
...
+	bys zipcode (cbsa): g missing_wgt = 1 - totwgt
+	bys zipcode (cbsa): egen temp_mis = count(res_ratio) if missing(res_ratio)
+	bys zipcode (cbsa): egen n_mis = sum(res_ratio==.)
+	g new_wgt = res_ratio
+	bys zipcode (cbsa): replace new_wgt = missing_wgt / n_mis if missing(res_ratio)
+
+	drop res_ratio totwgt missing_wgt n_mis
+
+	collapse (mean) safmr* [w = new_wgt], by(zipcode)
+end
```
diegogentilepassaro, line 237: This should all be gone! We should only care about the SAFMR zipcodes, match those with the geo master and obtain the county, state, and whatever geo variables.

derived/min_wage/code/make_mw_panel_zip.R
```diff
@@ -133,5 +158,43 @@ assemble_mw_US <- function(mxwalk, stmw, ctymw, locmw) {
   return(dfzip)
 }
 
+assemble_expmw_state <- function(x, p, zip, input) {
```
diegogentilepassaro, line 161: Let's remember [this comment](https://github.com/diegogentilepassaro/min_wage_rent/issues/92#issuecomment-781045450)!

diegogentilepassaro (changes requested):
Thanks for the awesome work @santiagohermo and @gabrieleborg, and sorry that it took me a lot to review! 

Overall things look great, but I left a few important comments. 
1. SAFMR looks like it needs a fix that uses the geo_master properly. 
2. @santiagohermo  pointed out that the current exp MW calculation does not take into account the inter-state commuters. We should fix that.

gabrieleborg:
@diegogentilepassaro the script already takes care of out-of-state workers, no need to change anything there!

derived/min_wage/code/make_mw_panel_zip.R
diegogentilepassaro: Thanks @gabrieleborg I'll resolve this comment then!

base/safmr/code/build_zipcode.do
santiagohermo: That makes sense!

base/safmr/code/build_zipcode.do
santiagohermo: Not sure. All zipcodes in the main US start above 1000?

diegogentilepassaro:
@santiagohermo and @gabrieleborg just did a pass to the raw SAFMR at base. 

Data from 2012 to 2016 comes at the zipcode-countyfips-cbsa10 level so I output a dataset at that level. For the years 2017 and on, data comes for zipcode-cbsa10 combinations only so I produced a different dataset for those years. At the derived stage, we should collapse those zipcode-some_geo_code combinations in the same way that we will do for the MW. 

The broad idea would be to have a dataset for MW for the whole US zipcode-county-cbsa10 combinations, as given by our geo_master and then collapse it to either the zipcode-month or the county-month using either house or population counts from geo_master. Then we will do the analogous for whatever other auxiliary dataset (QCEW, SAFMR, census), and finally we will merge Zillow data at the appropriate level. Does that make sense? Let's decide a criterion and path to do that! 

Also, could someone quickly review my changes in `base\safmr`?

santiagohermo:
Thanks @diegogentilepassaro! I cleaned up a bit the main program and re-run it. It looks much better than before!

santiagohermo:
I'm applying some parallelization in `lodes/make_odmatrix_ziplevel.R`. [This](https://dept.stat.lsa.umich.edu/~jerrick/courses/stat701/notes/parallel.html) is the source where I'm learning from, just for reference. FYI @diegogentilepassaro

santiagohermo:
Success @diegogentilepassaro @gabrieleborg. We now parallelize the computations from `base/lodes/code/make_odmatrix_ziplevel.R`. A few comments:

- The [code](https://github.com/diegogentilepassaro/min_wage_rent/blob/1832ec5a963be202d161da7c46ab66b880ee753a/base/lodes/code/make_odmatrix_ziplevel.R) is set for Windows. However, I _think_ that changing the line below should make it work for Mac
https://github.com/diegogentilepassaro/min_wage_rent/blob/1832ec5a963be202d161da7c46ab66b880ee753a/base/lodes/code/make_odmatrix_ziplevel.R#L41
- I ran it with [12 cores](https://github.com/diegogentilepassaro/min_wage_rent/blob/1832ec5a963be202d161da7c46ab66b880ee753a/base/lodes/code/make_odmatrix_ziplevel.R#L14), and it used +30gb of RAM and ~90% of CPU usage. Of course, if you are not in the server changed that
- Massive improvement in speed! Take a look at [the diff of the `make.log`](https://github.com/diegogentilepassaro/min_wage_rent/pull/94/commits/1832ec5a963be202d161da7c46ab66b880ee753a#diff-380d4d0bd9b87f6cae4c1b77ec42992003ff0101bfc3c13bf4ce5638d24a8ecd). Last time, it took a bit less than 4 hours. The run after updates in the server took 35 minutes.

Since [the data_file_manifest shows no differences in the data](https://github.com/diegogentilepassaro/min_wage_rent/pull/94/commits/1832ec5a963be202d161da7c46ab66b880ee753a#diff-c6a90fb5e4fde7929fddbc2bcb06b6db636987e18029c37c279198267152a309), I'll proceed to `squash and merge`!

diegogentilepassaro:
Genio y figura @santiagohermo!

PR #98, opened 2021-03-08
PR for #97update base/demographics

gabrieleborg:
@santiagohermo @diegogentilepassaro I have updated the `base/demographics` folder. I am addressing the crosswalk issue in a separate branch, so here you won't see any update in the way we use crosswalks. 

Following the main goal stated [here](https://github.com/diegogentilepassaro/min_wage_rent/issues/85#issuecomment-789999525), I have done the following: 

- I have removed two scripts: `tract_avgMW_13_18.R` and `make_MWrenters_ziplevel.R`. These were the scripts used to identify MW workers in the old way (i.e. using ACS data as opposed to lodes). We have stopped using the variables created with those; also in the heterogeneity section of the paper, we ended up using LODES to proxy for MW workers. 
- In order to create county-level demographics, I have first downloaded the same Census tables we used at the tract-level, but at the county level, and then I created a new, county-level, script called `make_demographics_countylevel.R`. This is essentially almost a copy of `make_demographics_ziplevel.R`, simply adapted to the new geographics level. 
- I have left the script `make_zippop_year.R`. This created a zipcode-month level panel data of the zipcode population, based on annual data (so all months in a year share the same observation). I remember that @diegogentilepassaro asked me for that when we were addressing Jesse's comments, but I couldn't find whether it was then used or not. You guys may want to have a look and let me know what do you want to do with it!

This PR will merge the branch into the pipeline branch #85.


base/demographics/code/make_demographics_ziplevel.R
```diff
@@ -17,7 +17,7 @@ main <- function() {
    table_list <- list.files(datadir, 
                             pattern = "*.csv")
    
-   table_list <- table_list[table_list != "nhgis0043_ds176_20105_2010_tract.csv"] # We are not using this file
+   table_list <- table_list[table_list != "nhgis0052_ds176_20105_2010_tract.csv"] # We are not using this file
```
diegogentilepassaro, line 20: Why did we change a file that we don't use? Are the changes in the data file manifest [here](https://github.com/diegogentilepassaro/min_wage_rent/pull/98/files#diff-c1986b226066735873253dba5ec003ebf76e3d27d3d02557daaf85e38394092a) for `zip_demo.csv` caused by this table change? Or the changes are due to the new geo master?

base/demographics/code/make_demographics_countylevel.R
```diff
@@ -0,0 +1,227 @@
...
+source("../../../lib/R/save_data.R")
+
+load_packages(c('tidyverse', 'data.table', 'tidycensus', 'bit64', 'readxl'))
+
+main <- function() {
+  
+  data_version <- "0056"
+  
+  datadir <- paste0("../../../drive/raw_data/census/county/nhgis", data_version, "_csv/")
+  xwalkdir <- "../../../raw/crosswalk/"
```
diegogentilepassaro, line 12: We don't seem to use this line. This demographics raw data comes already at the county level so that's why we don't use the geo master?

base/demographics/code/make_demographics_countylevel.R
```diff
@@ -0,0 +1,227 @@
...
+                             (unemp_share20105 / unemp_share20105D), 
+                             (employee_share20105 / employee_share20105D), 
+                             (worker_foodservice20105/worker_foodservice20105D))]
+  
+  denom_cols <- colnames(table_final)[str_detect(colnames(table_final), "D$")]
+  
+  table_final[, (denom_cols):= NULL]
+  
+  save_data(table_final, key = c('county_fips'),
+            filename = paste0(outdir, 'county_demo.csv'),
```
diegogentilepassaro, line 66: Probably we should name the files with the date that they belong to: `county_demo_2010.csv` and `zip_demo_2010.csv` maybe?

diegogentilepassaro (commented):
@gabrieleborg overall this task looks great! Thanks for the work :D

I posed a few quick questions and once I understand fully we can go ahead and merge if things looks right!

base/demographics/code/make_demographics_ziplevel.R
gabrieleborg: I updated the nhgis version because otherwise the code breaks. We ended up not using that file, so we remove its name from the file list, but since we are currently using version 0052, if we ask to remove `nhgis0043...` the code is going to break.
The differences you pointed out come from the updated tract-zip crosswalk!

base/demographics/code/make_demographics_countylevel.R
gabrieleborg: nice catch! I deleted it

base/demographics/code/make_demographics_countylevel.R
gabrieleborg: names updated

gabrieleborg:
@diegogentilepassaro thanks for the comments! I answered to each of those. Let me know if it looks good to you and when we can merge

diegogentilepassaro:
Thanks @gabrieleborg! The only remaining point is that I believe that given that we use the tract-zipcode crosswalk here to construct the `zip_demo_2010.csv' we will need to update it in the same fashion that we decided to do with LODES yesterday in #101. Do you agree with that? What do you think @santiagohermo ?

gabrieleborg:
I am not sure @diegogentilepassaro @santiagohermo. What problems may arise if we leave it like that? so far we are using demographics when building the final panel. Once we have a zipcode level demographics (for the whole US) we just merge that to our panel of selected zipcodes. Am I missing something?

santiagohermo:
Thanks @gabrieleborg @diegogentilepassaro! That's a subtle point.

Maybe, for consistency, we should construct the demographics at the ZCTA level and then replace the ZCTA by our preferred ZIP code? That way we assign all the people in the ZCTA to that ZIP code, as we do in the LODES construction. Is that what you have in mind @diegogentilepassaro?

However, I'm not sure we want to do this here, since we use this data mostly for heterogeneity analysis and stuff like that, and for that I believe we want to have the best characterization of the ZIP code possible, right? In the OD matrix this seems more important because otherwise we introduce missings when we match it with the statutory minimum wages.

diegogentilepassaro:
Thanks for your input @gabrieleborg and @santiagohermo! 

Seems that we can keep it as is for now and we can decide to change it if we change our minds and #101 goes well :D

Green light for merging!

gabrieleborg:
Thanks for the comments @diegogentilepassaro @santiagohermo :) 

If we ever decide to use ZCTA demographics, an alternative would be to directly download census data at that geography, hence avoiding using any crosswalk.

diegogentilepassaro (approved):

diegogentilepassaro (approved):

Issue #96, opened 2021-03-07
Update geo_master crosswalk (format tract_zip)


gabrieleborg:
**Goal**
Create a stable version of the crosswalk also for geographies smaller than zipcode (i.e. tracts). 

**Steps**
- format tract_zip crosswalk in `base/geo_master`
- update `base` subfolders by using the updated crosswalk

gabrieleborg:
Thread continues in #99

santiagohermo:
Summary [here](https://github.com/diegogentilepassaro/min_wage_rent/issues/85#issuecomment-801214783)

PR #99, opened 2021-03-09
PR for #96 update geo master

gabrieleborg:
Hey @santiagohermo @diegogentilepassaro ! 

I have reviewed `base` folders and updated scripts by homogenizing the adoption of crosswalks. Specifically: 

- I have modified [`base/geo_master/code/build.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/%2396-update-geo-master/base/geo_master/code/build.do) so to now create and format 2 main crosswalks: the one that Diego created, dealing with all geographies larger than zip codes, and a new one that deals with geographies smaller than zip codes (namely, census tracts). 
- I have updated scripts in the various `base/` folders to now import almost only these two crosswalks (the exception being the LODES-specific crosswalks, that are handled "internally"). Therefore now, instead of loading the tract_zip raw xwalk and formatting it every time we use it, scripts simply load the already formatted version from `geo_master/output`. 
- I have homogenized the choice of weights the 2 crosswalks use. Both crosswalks now use the number of houses as weight. 
- I have updated the REAMDE.d files containing dependencies. 

I have re-run `base/demographics` and it works fine. However, I wanted to ask if you can re-run `base/lodes/` since it's going to take a looong time on my laptop, while you can do it on the server :) 

This branch will merge on #85.


santiagohermo:
Thanks @gabrieleborg! I just set `base/lodes` to run in the server on this branch. I'll go through the code now.

santiagohermo (approved):
Thanks for the great work here @gabrieleborg! It looks to me that we are dropping a bunch of lines in `demographics` and `lodes`, so this pull makes me happy haha.

I approve. I had some questions about the code and data manifests that we can discuss in in-line comments. 

Once the run of `lodes` finishes in the server, I'll commit the changes to the branch and we can merge :)

santiagohermo:
The geomaster should have one zipcode per zcta. We will select one zip code per zcta with the following priority:

1. ZIP code shows up in Zillow data
2. ZIP code has maximum number of valid rents data
3. ZIP code is urban
4. Random

---

Steps:
1. aggregate zillow data to zipcode counting number of months
2. take zipcode to zcta xwalk and match data from previous step, thus adding: 
   - indicator for zipcode in Zillow data
   - number of months of valid sfcc rents data
3. Add other geographies based on zcta county place etc, keeping the count of houses.
4. Keep one zipcode per zcta following criteria above

Minor detail: can you keep two ZIP codes when both show up in Zillow data? How many such cases are there? If very few, ignore.

diegogentilepassaro:
@santiagohermo just in case take a look but I think that we are ready to squash and merge this one :D
Finally!!!!

santiagohermo:
Thanks @diegogentilepassaro! I went through the code and it makes sense to me. Hopefully, this is it! I'll go ahead and merge

fyi @gabrieleborg

PR #100, opened 2021-03-17
PR for #92: Build MW variables in `derived/min_wage`

santiagohermo:
The build in #92 is completed @diegogentilepassaro @gabrieleborg! Now `derived/min_wage` is organized as follows:

- `build_statutory.R` takes the geo_master and the MW data from `base` and builds datasets at the ZIP code and county leves at a monthly frequency with the (actual) statutory MW, saved in `drive/output/derived_large`. Per discussion with @diegogentilepassaro, we also keep rural places here.
- build_experienced.R` takes the dataset at the Zip code level produced above and the OD LODES matrices, and constructs a new dataset with the experienced MW, saved also in `drive/output/derived_large`. With parallelization over states

Both datasets are also saved in `.dta` format for easier assembly later on. Let me know what you think!


santiagohermo:
Per "in-person review" in our meeting today I will squash-and-merge this PR. Thanks @diegogentilepassaro @gabrieleborg!

Issue #101, opened 2021-03-17
A new fix for LODES origin-destination matrices


santiagohermo:
In this issue we will update the construction of the LODES OD matrices.

First, we will set the "destination threshold" to 1.

Second, we will fix the construction following these steps:
1. Aggregate from block-group level to tract level as now
2. Aggregate from tract level to ZCTA (instead of ZIP code)
3. Assign a ZIP code to each ZCTA based on the `geo_master`

Finally, as a sanity check, _assert_ that number of workers in home equals sum of number of workers in each workplace.

**_UPDATE MARCH 24_**

**GOAL**
Update geo-master construction so to have a zipcode-place-county level frame that can be then used as template for other datasets.

**STEPS**
1. Update geo-master: return a zipcode-place-county level dataset
2. update OD matrix construction 
    - Aggregate from block-group to tract as now
    - Aggregate from tract to zipcode (as now)
    - set destination threshold to 1 
3. Update MW construction, using new geo-master template

gabrieleborg:
@diegogentilepassaro @santiagohermo I am working on this. But I keep forgetting what was the main issue here. Can you guys write it here? so we have it recorded :) thanks!

santiagohermo:
> @diegogentilepassaro @santiagohermo I am working on this. But I keep forgetting what was the main issue here. Can you guys write it here? so we have it recorded :) thanks!

The main issue here is that, since in geo_master we keep one zip code per zcta, when we match the LODES to our crosswalk we lose ZIP codes (the ones we dropped in the crosswalk). Thus, we have missing MWs in the OD matrix, which means that some commuters are in practice being dropped out of the computations.

To avoid this, the construction of the OD matrix should be at the zcta level, and then we "replace" the ZCTA by our "chosen" zip code. That way we won't have any missing when matching the OD matrix to the MWs, since these are constructed on the geo_master. The assumption we will be making is that all the people in a zcta reside in the zip code that we select for that zcta.

Do you agree @diegogentilepassaro @gabrieleborg?

diegogentilepassaro:
Agreed @santiagohermo! Thanks for writing it down very clearly 👍

gabrieleborg:
thank you @santiagohermo !

gabrieleborg:
Hey @santiagohermo @diegogentilepassaro! 

[...]

PR #103, opened 2021-03-26
PR for #101: More thoughts on geo_master + re-run lodes OD matrices

santiagohermo:
PR for #101. Per [here](https://github.com/diegogentilepassaro/min_wage_rent/issues/101#issuecomment-808328114), there won't be review here.


diegogentilepassaro:
❤️

gabrieleborg:
Thanks @santiagohermo !

About the increased number of origin-destination combinations, yeah we previously cut a loooong right tail of the distribution!

Issue #102, opened 2021-03-22
Update zipcode-month panel in `derived`


diegogentilepassaro:
Part of #85

[Goal]
Create a new updated zip code-month panel in `derived'

[Steps]
1. Use the new `base' level directories to create a zip code-month panel that allows us to estimate our baseline models. It needs to contain MWs, experiences MWs, rents, economic controls, and demographics. Keep all zip codes, regardless of whether they have rents data or not as that can be useful for assessing the representativeness of our data in terms of zip codes, but also in terms of what share of the MW changes we cover in our exercises.
2. Compare the number of zip codes, the number of zip code-month, and the number of treated zip codes with before the change in pipeline. Did we gain zip codes? 
3. Sanity check that we have MW data for the whole US in our 2010-2019 period. Create some maps. Compute how many events we use and how many there are at the state and local level. Also. what is the fraction of zip codes that experience a change in MW in the whole data and what is that number in our data that has rentals?

diegogentilepassaro:
@gabrieleborg and @santiagohermo, I have created branch `issue102` to work in this task. I wanted to mention that I will use as a base panel the MW one, that should be for the whole US, and the I will merge the others to that one. Does that make sense to you?

diegogentilepassaro:
A few initial notes @santiagohermo and @gabrieleborg : 
- Even if we start from the MW derived panel, we still need to join to our geo master to get other geographies than zip code. This is relevant for FEs and clustering.
- The `year_month` variable living in `../../../drive/derived_large/min_wage/zip_statutory_mw.csv` is not imported properly to STATA. We probably should make sure we use the proper options of fwrite for dates.
- The `zipcode` variable does not have leading zeros when zip codes start with 0. We should probably fix that by always saving them as a string.

diegogentilepassaro:
@santiagohermo and @gabrieleborg, I started addressing [this comment](https://github.com/diegogentilepassaro/min_wage_rent/issues/102#issuecomment-805351420) and implementing the task. A few new issues that I encountered:

1. When merging the experienced MW panel to the baseline ZIPcode-year-month statutory MW panel we fail to match a lot of observations. It seems that this may get fixed when we have a working geo master, but flagging it in case you think of another reason that I should be aware of.
![Screen Shot 2021-03-24 at 6 33 11 PM](https://user-images.githubusercontent.com/20582494/112403536-f6718900-8ce4-11eb-8f89-a9a8c9fa3345.png)

[...]

PR #104, opened 2021-04-07
PR for #102: Update derived zipcode-month and county-month panels

### LINKED ISSUE #85: Update pipeline

diegogentilepassaro:
[GOAL]
Review and, if needed, improve the construction of zipcode-month and county-month panels.

[STEPS]
1. Quickly review the changes in the construction of the geo master. The goal there is to have:
    - A dataset the has for every USPS zipcode **all** possible county-place combinations.
    - A dataset that has for every USPS zipcode a unique county-place combinations that is chosen based on the one that has more houses, and in the case of a tie it breaks it randomly.
    - A tract to USPS zipcode dataset. Please flag where we use this dataset.
    - A tract to ZCTA dataset. Please flag where we use this dataset.
2. Quickly review the changes in `base`. The goals here are:
    - Make sure we always use the geography that comes in each raw data, and if we are going to apply some transformation of those it has to be through some dataset in the geo master. 
    - Make sure that geo ids at base are always strings and that whenever there is a time dimension we create numeric variables as detailed as possible. For example, if the data is at the year month, instead of creating year-month let's create a numeric variable for year and another for month. These steps eases merging across datasets.
    - Make sure that the way we create LODES data makes sense and let's discuss adding a county level experienced MW so that we can replicate our model at the county-month and compare it to Yamagishi and Tidemann in better shape. 
3. Review changes in `derived`. This is the bulk of the work! Here we need to do the following:
    - Quickly review the `min_wage` directory.
    - Review the construction of the zipcode-month panel. Let's open a thread on the zipcode-months that don't merge and discuss the different cases as per [your comment here](https://github.com/diegogentilepassaro/min_wage_rent/issues/102#issuecomment-811778123). As I told you today, we don't really have lose Zillow zipcodes but we do have a few strange cases still, for example, zipcode-months with experienced MW and not statutory.
    - Review the construction of the zipcode level dataset. This should be super easy. Mainly let's make sure that I didn't forgot to add some useful zipcode level data.
    - Review the construction of the county-month panel. Again, as in the zipcode-month panel, let's be careful about discussing the cases for the merges. I believe that we do lose some counties and that is a little weird. 
    - Review the construction of the county level dataset. Should be extremely easy.

After this PR we can finally close #85 and start the build of the estimating panels, the descriptive stats and maps, and the new analysis as per the plan on the [updated outline](https://github.com/diegogentilepassaro/min_wage_rent/wiki/Outline!!!!!)


diegogentilepassaro:
FYI @santiagohermo!

gabrieleborg:
### STEP 1: Review of `base/geo_master`

- `zip_county_place_usps_all` and `zip_county_place_usps_master`: ok
- `tract_zip_master`: crosswalk used in
    1. [`base/demographics/code/make_demographics_ziplevel.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/demographics/code/make_demographics_ziplevel.R)
    2. [`base/demographics/code/make_zippop_year.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/demographics/code/make_zippop_year.R)
    3. [`base/lodes/code/make_xwalk.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/lodes/code/make_xwalk.R)
    4. [`base/lodes/code/make_lodes.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/lodes/code/make_lodes.R) but unnecessary (see upcoming review of `base/lodes/`)   
- `tract_zcta_master`: this crosswalk is never used neither in `base` nor in `derived`. So technically we can delete the following lines: 
    https://github.com/diegogentilepassaro/min_wage_rent/blob/8a4cb0f79e55732b435acc492da9f254930ce01a/base/geo_master/code/build.do#L78-L95

gabrieleborg:
### STEP 2: Review of `base`

- #### Key variables (geography and time), updates needed:
    1. [`base/bps/code/make_bps_cty_month.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/bps/code/make_bps_cty_month.do): 
        1. convert to string geography (countyfips)
        2. update time variable
        3. NOTE: we don't use this piece of data anymore. Shall we keep it just to be safe, and quickly update it? 
    2. [`base/demographics/code/make_demographics_ziplevel.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/demographics/code/make_demographics_ziplevel.R): 
        1.  convert to string geography (zipcode)
    3. [`base/demographics/code/make_demographics_countylevel.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/demographics/code/make_demographics_countylevel.R):
        1. convert to string geography (countyfips)
    4.  [`base/demographics/code/make_zippopo_year.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/demographics/code/make_zippop_year.R):
         1. convert to string geography (zipcode)
     5. [`base/lodes/code/make_lodes.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/lodes/code/make_lodes.R):
        1. convert to string geography (zipcode)
    6. [`base/lodes/code/make_odmatrix.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/lodes/code/make_odmatrix.R):
        1. convert to string geography (h_zipcode, w_zipcode) 
   7. [`base/min_wage/code/state_mw.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/min_wage/code/state_mw.do):
       1. convert to string geography      (statefips)
       2. split time variable (there are 4 different output here, daily, monthly, quarterly, and annually. So each needs a specific new set of time variables)
   8. [`base/min_wage/code/substate_mw.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/min_wage/code/substate_mw.do):
        1.  split time variable (there are 4 different output here, daily, monthly, quarterly, and annually. So each needs a specific new set of time variables)
    9. [`base/qcew/code/create_ind_county_quarter_panel.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/qcew/code/create_ind_county_quarter_panel.do):
        1. split time variable (year_quarter)
    10. [`base/qcew/code/make_county_month_panel.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/qcew/code/make_county_month_panel.do):
        1. split time variable (year_month)
    11. [`base/zillow/code/build_zipcode.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/zillow/code/build_zipcode.R):
        1. convert to string geography (zipcode)

- #### Review LODES and steps to build county-level lodes:
    1. I reviewed once again the construction of the [od matrix](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/lodes/code/make_odmatrix.R), and it makes sense to me.
    2.  If we decide to build a county-level experienced MW, this should be hard. Specifically, we could create a very similar script to `base/lodes/code/make_odmatrix.R`, where we aggregate workers' flows at the county-level instead that at the zipcode level. To do that is even simpler, since we won't need any geo master but only the LODES specific crosswalks that allocate each block group to all census geographies (but the zip codes of course, and that's why we needed the small xwalk in our main case). We could simply add a function in [`base/lodes/code/make_xwalk.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/lodes/code/make_xwalk.R) similar to `make_xwalk_od` (reported here for visual clarity), and use that function to aggregate the odmatrix at the county level. Then in `derived` we would simply need to compute the experienced MW using such matrix.
    https://github.com/diegogentilepassaro/min_wage_rent/blob/8a4cb0f79e55732b435acc492da9f254930ce01a/base/lodes/code/make_xwalk.R#L2-L19

gabrieleborg:
@diegogentilepassaro added a first pass of the review. Since I cannot work on the files, I reviewed them here on GIthub and listed all changes we need to implement and their location in the code. Let me know if it makes sense!

diegogentilepassaro:
@gabrieleborg, I just implemented the changes for Step 1 that [you proposed](https://github.com/diegogentilepassaro/min_wage_rent/pull/104#issuecomment-815590205). I will re run the dependencies once we also implement the updates as per Step 2.

diegogentilepassaro:
I started going with [the changes](https://github.com/diegogentilepassaro/min_wage_rent/pull/104#issuecomment-815633804) for Step 2:

- #### Key variables (geography and time), updates needed:
    1. [`base/bps/code/make_bps_cty_month.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/bps/code/make_bps_cty_month.do): 
       -  Let's keep bps data. I fixed the geo and time vars and I re run.
    2. [`base/demographics/code/make_demographics_ziplevel.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/demographics/code/make_demographics_ziplevel.R): 
        -  I changed the geo vars to strings. It was a little complicated! :D 
        - One question: in line 20 we say that we don't use some file. Can we expand the comment to explain why? 
        - Can we change the data version to 0055 so that is the same as in [`base/demographics/code/make_zippopo_year.R`]. Does that make sense? Why we have different versions? Can we use the same version across the county file too?
    3. [`base/demographics/code/make_demographics_countylevel.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/demographics/code/make_demographics_countylevel.R):
        -  I changed the geo vars to strings. Same question: in line 19 we say that we don't use some file. Can we expand the comment to explain why? 
    4.  [`base/demographics/code/make_zippopo_year.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/demographics/code/make_zippop_year.R):
         - Converted geo vars to string. The year was fixed as part of #102. Why we don't eliminate the file from NHGIS that we don't use before here?

A big picture comment on the demographic directory: is it possible to avoid hard coding of the NHGIS files?

gabrieleborg:
Thanks @diegogentilepassaro for taking care of the changes!! Here are my replies regarding **STEP 2**:

- [`base/demographics/code/make_demographics_ziplevel.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/demographics/code/make_demographics_ziplevel.R):
    -   _In line 20 we say that we don't use some file. Can we expand the comment to explain why?_
        The reason is that in the original download we included some variables that we ended up not using. These variables come in a different census' table, and since different tables come in different files, we have a csv we don't use. Since we list all csv using the R function `list.files`, we need to exclude that table from the list, otherwise the code breaks downstream. 
    - _Can we change the data version to 0055 so that is the same as in `base/demographics/code/make_zippopo_year.R]`. Does that make sense? Why we have different versions? Can we use the same version across the county file too?_
    No we cannot if we want to keep these downloads. The reason is that data version corresponds to different downloads. So right now version 0055, for example, only contains population data from different years (to build the yearly population), while the version we use to make zip code demographics contains many more variables, but only from 2010 Census and 2008-2010 ACS (or something like that). Similarly, to build county-level demographics we use a different version (i.e. download) where we have the same variables but at county-level. If we want to homogeneize the NHGIS version, what we could do is the following: 
        1. Make an exhaustive list of all tables - with the corresponding geographic level - that we use across all scripts in `base/demographics`. We can do this also by simply looking at the variables we use (since variables at a specific geo level are found in a specific table). 
        2. make a new NHGIS download with **all** tables/variables together (so, tract-level demos from 2010 and 2010 ACS, county-level demos from 2010 Census and 2010 ACS, tract-level population from ACS from multiple years). This download will have a new version number (depending on who does it, because the version number is simply the number of downloads made by an NHGIS account. So, for example, version 0055 means that that was my 55th download with my account). 
        3. Update all scripts in `base/demographics` to select the right tables. This implicitly means that we will have to make the script a bit more complex because in this case we cannot list all files (tables) for that version. We will have to manually list all tables we need in each substep. 
        
        So to conclude, @diegogentilepassaro let me know if that is clear and what do you think it's best to do in this situation! Given that I don't have much time I would say leave it like that, but if you think you can manage it, I think that the previous points I listed show a "relatively quick" way to address your concern!

- [`base/demographics/code/make_demographics_countylevel.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/demographics/code/make_demographics_countylevel.R): 
    -  _Same question: in line 19 we say that we don't use some file. Can we expand the comment to explain why?_ 
    Here we dragged the same issue from above because, to replicate the same exact demo variables we had a zip level, I only modified my NHGIS download by changing the geography I was using but I didn't touch the variables selected. So the extra table appeared in the county-level version/download too. 
- [`base/demographics/code/make_zippop_year.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/demographics/code/make_zippop_year.R):
    -   _Why we don't eliminate the file from NHGIS that we don't use before here?_ 
    I am not sure I understand this question. Can you elaborate on this? 


I hope my answers were clear enough! To address your general picture question, I think I partly answered in my previous points. I think there's a trade-off in terms of hard coding with NHGIS data. If we want to use only one version (one download with all tables at all geographics levels), then we will need to manually select tables in each script. While if we keep different version for different geo (and time) levels, then we only need to "hard-code" the version of each download in the corresponding scripts and then use the `list.files` function to retireve them all.

gabrieleborg:
### STEP 3: Review of `Derived`

- [`derived/min_wage/code/build_statutory.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/derived/min_wage/code/build_statutory.R): OK 
- [`derived/min_wage/code/build_experienced.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/derived/min_wage/code/build_experienced.R): 
    - update the time variables
- [`derived/zipcode/code/build_zipcode_panel.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/derived/zipcode/code/build_zipcode_panel.do):
    - convert to string geography (zipcode)
- [`derived/county/code/build_county_panel.do` ](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/derived/county/code/build_county_panel.do):
    - convert to string geography (countyfips)
- [`derived/zipcode_month/code/build_zipcode_month_panel.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/derived/zipcode_month/code/build_zipcode_month_panel.do):
    - convert to string geography (zip code) 
    - The code looks very clear now, thanks! just to reiterate: we are not merging together zipcode and zipcode_month dataset here because we are going to do it in specific `analysis`' subtasks, right (i.e. in a `preclean` script only if we need, say, to add demographics)?     
    - I will open a specific issue on the "unmatched" zip codes after I post this. 
- [`derived/county_month/code/build_county_month_panel.do`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/derived/county_month/code/build_county_month_panel.do):
    - The script looks fine to me, but I still couldn't figure out why we lose some counties. I'll keep thinking about it for a while. 
    - I also agree that we should build a county-level experienced MW, as well as population variable.  For the former, see [this previous post](https://github.com/diegogentilepassaro/min_wage_rent/pull/104#issuecomment-815633804). For the latter, [this reply](https://github.com/diegogentilepassaro/min_wage_rent/pull/104#issuecomment-816481587).

gabrieleborg:
Hey @diegogentilepassaro ! 

I made the first pass in creating the county-level experienced MW. 

1. I created [`base/lodes/code/make_odmatrix_county.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/lodes/code/make_odmatrix_county.R). This is done by modifying [`base/lodes/code/make_odmatrix.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/lodes/code/make_odmatrix.R), and aggretating both home and workplace geographies at the county-level. Note that to do that, I added the `make_xwalk_rac_wac_county.R` function in the [`base/lodes/code/make_xwalk.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/base/lodes/code/make_xwalk.R) file.
    - also note my inline comment at `line 60`.   
2. Similarly, I then created [`derived/min_wage/code/build_experienced_county.R`](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_102/derived/min_wage/code/build_experienced_county.R), where I import the newly created county-level od matrices and , also using `drive/derived_large/min_wage/count_statutory_mw.csv`, assemble the `drive/derived_large/output/county_experienced_mw.csv`.  


As discussed yesterday, I wasn't able to run this commit but if you do, please let me know whether any error comes up! thank you :) 🙏

diegogentilepassaro:
@gabrieleborg I am through with all the changes in base and @santiagohermo and I iterated on in derived/min_wage and now is complete as well!

Now the only pending thing is to do the checking of the merges in derived zipcode-month, zipcode, county-month, and county. Let us know if you have some time this week so that we agree on this stage and move on :D

gabrieleborg:
Hey @diegogentilepassaro @santiagohermo ! 

Thanks for the update. I will try to have a look between this afternoon and tomorrow morning. I will update you tomorrow around lunchtime.

gabrieleborg:
@santiagohermo @diegogentilepassaro In order to start reviewing `derived/zipcode_month` and `derived/county_month` I had to re-run `base/lodes` and `derived/min_wage`. They are still running so I'll try to finish this afternoon.

santiagohermo:
> @santiagohermo @diegogentilepassaro In order to start reviewing `derived/zipcode_month` and `derived/county_month` I had to re-run `base/lodes` and `derived/min_wage`. They are still running so I'll try to finish this afternoon.

Thanks @gabrieleborg! Note that we now save output from those steps in GDrive. If it's not updated, then we should probably update it. fyi @diegogentilepassaro

Issue #111, opened 2021-09-04
Develop main function to estimate model + update main figures


santiagohermo:
This issue is tracked by #107. In here, we will:

1. Write a function that estimates the distributed lag model and saves results to a csv or a tempfile or leaves them in memory. We want the function to be flexible so that we can run many estimations in loop and append all results in the same dataset.
2. Use function to update `analysis/first_differences`, which should now produce table 1 for static model and main dynamic model figures, following [the slides](https://github.com/diegogentilepassaro/min_wage_rent/blob/61ef6f1b3314b758f6d606fd84d353caf92b77e1/paper_slides/slides/long_slides.pdf)

Issue #113 will update other estimations in `analysis` following the structure developed here.

diegogentilepassaro:
Starting here @santiagohermo!

diegogentilepassaro:
Created branch `issue_111` and started out with some housekeeping. I deleted:

- `analysis/estimates_benchmark'
-  `analysis/event_size_robustness'
-  `analysis/estimates_benchmark'
-  `analysis/estimates_benchmark'
-  `analysis/estimates_benchmark'
-  `analysis/estimates_benchmark'

santiagohermo:
> Notes:
> 
> * Opened branch `#111_main_results` for related work

Fijate que ya había una branch abierto querido @diegogentilepassaro. La podes borrar?

diegogentilepassaro:
Hola @santiagohermo! No veo ninguna:
<img width="1399" alt="Screen Shot 2021-09-07 at 11 25 28 AM" src="https://user-images.githubusercontent.com/20582494/132371057-324c6b04-cfe4-4735-baee-fe4171842e32.png">

santiagohermo:
Oops, gracias por chequear @diegogentilepassaro! Parece que nunca la publiqué (quedó local en mi machine nomás). Borré esa nota del [opening comment](https://github.com/diegogentilepassaro/min_wage_rent/issues/111#issue-988203445).

diegogentilepassaro:
[...]

Issue #113, opened 2021-09-04
Update all estimations


santiagohermo:
This task is tracked by #107. A new code structure for estimating the models will be developed in #109. That structure will be implemented only in the main estimations (to produce static table and main dynamic figures).

In this task we will implement the new code structure in all estimations, which should include:
- Unbalanced, reweighted and unbalanced reweighted models
- Arellano-Bond models, to be included fully in an appendix

This task should start after #109 is completed

diegogentilepassaro:
@gabrieleborg with the tools developed in #111 this task should be self-contained. Let us know if you think that you can take care of it or you want to split it in some convenient way :D

PR #117, opened 2021-09-08
PR for #111: build function for estimating distributed lag models cleanly and re do main tables and figures

diegogentilepassaro:
Follow up for #111 

[Goal]
Make sure that our main analysis is now cleanly produced and correct. 

[Steps]
1. Review comment thread of #111.
2. Make sure that `estimate_dist_lag_model` is correct and handles all tested cases without issues.
3. Review `analysis\fd_baseline\estimates.do` and make sure that the production of the inputs for creating figures and tables look good.
4. Review `analysis\fd_baseline\figures.do` and make sure that figures are plotted correctly and that the functions that create them are flexible to handle future models (think of models that we have tried in the past: levels, AB, etc)
5. Review and understand `analysis\tables`.


diegogentilepassaro:
Continues in #117!

analysis/fd_baseline/code/tables.R
```diff
@@ -0,0 +1,55 @@
+remove(list = ls())
+
+
+main <- function() {
+  
+  instub  <- "../output/"
+  outstub <- "../output/"
+  
+  est <- read.csv(file.path(instub, "estimates_static.csv"))
```
gabrieleborg, line 9: Is there any reason for why we only re-format static estimates? Maybe we would want to add dynamic tables in appendix? (in this case, we could only keep the final cumsum so not to make the table huge, given that figures will show how the cumulative sum is essentially stable).

analysis/tables/code/make.py
```diff
@@ -25,16 +26,13 @@
...
-    stata = "StataMP-64"
-
 start_make_logging()
 
-run_stata(program = 'figures.do', executable = stata)
+fd_baseline_dir = "../../fd_baseline/output"
+
+tablefill(input    = fd_baseline_dir + '/static.txt', 
+	      template = '../input/static.tex', 
+          output   = '../output/static.tex')
```
gabrieleborg, line 35: Same comment as for `fd_baseline/table.R`: is there any reason why we aren't producing tables reporting dynamic estimates?

analysis/tables/input/static.tex
```diff
@@ -0,0 +1,29 @@
...
+     & \multicolumn{1}{c}{(3)} & \multicolumn{1}{c}{(4)}                         \\ \midrule
+    $\Delta \ln \underline{w}_{ict}$          &  #4#  &  #4#  &       &  #4#     \\
+                                              & (#4#) & (#4#) &       & (#4#)    \\
+    $\Delta \underline{w}_{ict}^{\text{exp}}$ &       &       &  #4#  & #4#      \\
+                                              &       &       & (#4#) & (#4#)    \\ \midrule
+    $\Delta \ln \underline{w}_{ict}+
+      \Delta \underline{w}_{ict}^{\text{exp}}$&       &       &       &  #4#     \\
+                                              &       &       &       & (#4#)    \\
+                                              &       &       &       &          \\ \midrule
+    County-quarter controls                   &  Yes  & Yes   & Yes   & Yes      \\
```
gabrieleborg, line 19: Maybe we want to make this flexible enough so to automatically output tables for any given geo-time aggregation level.

analysis/tables/input/static.tex
```diff
@@ -0,0 +1,29 @@
...
+    P-value equality                          &       &       &       & #4#      \\
+    R-squared                                 &  #4#  &  #4#  &  #4#  & #4#      \\
+    Observations                              & #0,#  & #0,#  & #0,#  & #0,#     \\\bottomrule
+    \end{tabular}
+
+    \begin{minipage}{.95\textwidth} \footnotesize
+        \vspace{2mm}
+        Notes: 
+    \end{minipage}
+\end{table}
```
gabrieleborg, line 29: The script returns a floating table. Are we sure we want that, as opposed to just the tabular? I remember we would fill titles and notes in the latex doc, as we drafted the paper. There could be pros in doing that, but open to go with this setting if you guys think it works better!

gabrieleborg (commented):
Hey @diegogentilepassaro @santiagohermo ! 

Finished reviewing this PR :) here are my general comments (for more specific ones, please see the inline ones):

- I think the updates are great. I was able to successfully run `analysis/fd_baseline` and get the desired outcome (although I had to do it manually as I would get a python error when try to run `make.py`, regarding the logs). 
- I checked the `estimate_dist_lag_model.ado`: I think it's flexible enough to accommodate robustness checks and future iterations. My only comment here is: 
    - line 26: any reason why the test for `F1D.dyn_`var'` is kept outside the loop ?
- `fd_baseline/estimates.do` and `fd_baseline/figures.do` looks good to me :) 
- I like how we then have an ad-hoc procedure to format output tables. I left a couple of specific comments inline. 
- I wasn't able to run the `analysis/table/` folder, but I studied the code: for what I understand first we take the estimates, then we re-format them using `fd_baseline/tables.R`, and then finally we put them in latex format. Is that correct? 

Besides this, we're good to go here!

diegogentilepassaro:
Thanks for reviewing @gabrieleborg :D 

This is super helpful. I am taking care of most comments but about some we will ping @santiagohermo as he constructed the static tables. Will reply soon!

analysis/fd_baseline/code/tables.R
diegogentilepassaro: This sounds like a good idea but I think it would be a great follow up task to add those tables, as they are not going to live in the main analysis. What do you think @santiagohermo ?

analysis/tables/code/make.py
diegogentilepassaro: @santiagohermo

analysis/tables/input/static.tex
diegogentilepassaro: Great idea, I think that would soon be important as we rewrite. @santiagohermo should we do that now or take care of that in a follow-up where we run and output things for the whole county-level analysis?

analysis/tables/input/static.tex
diegogentilepassaro: Both approaches should work because we can change the titles and notes in static.tex and then input that into the draft. @santiagohermo, what do you think?

diegogentilepassaro:
@gabrieleborg I moved line 26 into the loop and things look the sam. We saved a line and code runs slightly faster. I also confirmed that `analysis/tables` runs for me. Once @santiagohermo gives his input we can decide how to move on and merge.

analysis/fd_baseline/code/tables.R
santiagohermo: Thanks @gabrieleborg! Absolutely, we can add more tables. The idea of the script was to produce only one to "illustrate" how it'll work using a single `csv` files with all estimation output. To speed-up this PR, I suggest we tackle the addition of tables in a follow-up issue (maybe #113).

analysis/tables/code/make.py
santiagohermo: See [here](https://github.com/diegogentilepassaro/min_wage_rent/pull/117#discussion_r707660656)

analysis/tables/input/static.tex
santiagohermo: Can you give more details about your proposal @gabrieleborg? I think that automating tables in STATA is a pain, so this new structure provides an alternative in which table "templates" are created manually, and then filled with the relevant numbers.

Let me be more clear on how the current process operates:
1. Take the `csv` file with the estimates of interest. Create a script that outputs a text file like [this one](https://github.com/diegogentilepassaro/min_wage_rent/blob/e5a85708629ad5d910953642c0d62dc654f6ed54/analysis/fd_baseline/output/static.txt), where the order in which the numbers appear matches how they'll be displayed in the table. 
2. Make a template like [this one](https://github.com/diegogentilepassaro/min_wage_rent/blob/e5a85708629ad5d910953642c0d62dc654f6ed54/analysis/tables/input/static.tex), where you specify the formatting of the numbers. A nice online tool to make new tables is [here](https://www.tablesgenerator.com/).
3. Fill the template using the text file as in [here](https://github.com/diegogentilepassaro/min_wage_rent/blob/d827e933937e16da5971b85cbb6f05866dffeb0a/analysis/tables/code/make.py#L33-L35).

To produce a different table we would need to create a new template and text file.

----
What alternatives do we have? I can think of 3.

1) Going back to STATA tables from `esttab`.
2) Keep this approach where table templates can be fully customized manually.
3) Fully create the table inside `R` using the csv with the estimates. This would be hard to automate through a single function though, as our tables typically look different.

My preference is: 2 > 1 > 3.

analysis/tables/input/static.tex
santiagohermo: I agree that both approaches should work @diegogentilepassaro @gabrieleborg. The issue here is that [`tablefill`](https://github.com/gslab-econ/gslab_python/tree/master/gslab_fill) fills a table with a label (I don't think you can fill a tabular). If we keep this approach, we would have to change notes and titles in the template of the table. While it's a disadvantage not being able to change this in the paper, the good thing is that all templates will live in the same place (`analysis/tables/input`).

Let me know what you think!

santiagohermo:
Thanks @diegogentilepassaro @gabrieleborg! I went through the open loops. [This comment](https://github.com/diegogentilepassaro/min_wage_rent/pull/117#discussion_r707677386) is probably the most useful one.

analysis/tables/input/static.tex
diegogentilepassaro: I vote for 2.!

analysis/tables/input/static.tex
diegogentilepassaro: I think this is not a big cost and adds a lot of flexibility so I would keep things as they are.

diegogentilepassaro:
We are almost ready to merge here @gabrieleborg and @santiagohermo! I wait for @gabrieleborg to answer and will squash and merge afterwards.

santiagohermo:
Per off-github discussion I'm closing this issue @gabrieleborg. Feel free to give your thoughts on the open points, and we can always make updates in #113!

santiagohermo:
Summary [here](https://github.com/diegogentilepassaro/min_wage_rent/issues/111#issuecomment-919573922)

Issue #277, opened 2024-06-26
Create `base/zcta_master`


santiagohermo:
In this issue we want to create a geography master of ZCTAs to be used with ACS data.

The idea is to create `base/zcta_master` following `base/zipcode_master` and make necessary changes to create the new geographies.

diegogentilepassaro:
@santiagohermo I am through with this one. Will send for a quick PR

PR #279, opened 2024-06-26
#277 Adding base/zcta_master

diegogentilepassaro:
Added base/zcta_master imitating base/zipcode_master


base/zcta_master/output/data_file_manifest.log
```diff
@@ -0,0 +1,32 @@
...
+n_census_b~s |     32,990    335.2858    1750.543          1     312601
+n_blockgro~s |     32,990    9.750773    23.21944          1       3748
+    n_tracts |     32,990     4.65923    15.97988          1       2797
+    n_places |     32,990    2.455289    5.601814          1        979
+-------------+---------------------------------------------------------
+  n_counties |     32,990    1.380691    4.533966          1        817
+      n_cbsa |     32,990    1.209306    2.064451          1        367
+    n_states |     32,990     1.00679    .2688513          1         48
+  num_houses |     32,990    3991.898    5550.697          0      47617
+  population |     32,990    9358.331    13667.13          0     113916
```
santiagohermo, line 19: Interesting to see that `zcta`s are a bit smaller! Compare to `zipcode_master` [here](https://github.com/diegogentilepassaro/min_wage_rent/blob/1f2abe5369a006c256a2dbbf90ad122a45f71086/base/zipcode_master/output/data_file_manifest.log)

base/zcta_master/code/build.do
```diff
@@ -0,0 +1,27 @@
...
+program main
+    local instub  "../../../drive/base_large/census_block_master"
+    local outstub "../../../drive/base_large/zcta_master"
+    local logfile "../output/data_file_manifest.log"
+
+    use `instub'/census_block_master.dta if !missing(zipcode), clear
+
+    make_baseline_frame, master_geo(zcta)
+
+    assign_geography, instub(`instub') master_geo(zcta) geo_to_assign(tract)
```
santiagohermo, line 15: que lindo cuando una función bien armadita hace su magia :heart:

santiagohermo (approved):
Looks great, thanks for the quick completion @diegogentilepassaro! 

We need to open up a new issue to continue pushing the zcta-level data forward.

base/zcta_master/output/data_file_manifest.log
diegogentilepassaro: I am now not 100% sure that should be the case! I am looking into [this directory](https://github.com/diegogentilepassaro/min_wage_rent/tree/master/base/census_block_master) and it seems that we get the mapping between census blocks and zcta from LODES. Could there be some missing there ?

diegogentilepassaro: Basically, what I wonder is if using the census blocks from LODES is granted to be the full set of census blocks or not! But I think it should be! When looking at a map of ZCTA there are some grey areas that are not covered, mostly national parks, airports, and unpopulated areas (see for example [here](https://peter-horton.com/2022/12/30/zip-codes-zctas-and-zctas-that-cross-state-boundaries/)). I guess that it makes sense then that they are less in that case, since USPS ZIP cover everything.

diegogentilepassaro: Just in case I wait for your thoughts to merge, but I think e should be fine :D

santiagohermo: As you say, we are using a mapping of census blocks to zctas from LODES. I think it makes to presume is correct!

Maybe zctas are sometimes created by splitting a zipcode that crosses a state, which is why there are slightly more zctas which are also a bit smaller?

In any case, I'll squash & merge here :)

Issue #276, opened 2024-06-26
Clean `acs_rents`


santiagohermo:
In this issue we want to clean the `acs_rents` data that was collected in #275.

The goal is to create `base/acs_rents_clean` with all the variables in a zcta-year panel.

santiagohermo:
Completed with exception of the variable "rent as share of income" @diegogentilepassaro. I'll go to PR in a bit, but I first wanted to explain the two issues with this variable.

### Issues with B25070

#### 1. Data gives counts per bin

The data in `B25070_gross_rent_as_a_percentage_of_household_income_block_group` looks like this:
![image](https://github.com/diegogentilepassaro/min_wage_rent/assets/45404755/2b211f5b-01b4-47c3-9167-ca5a10de9d3a)

<details>
  <summary>Metadata</summary>
  
```csv
"Column Name","Label"
"GEO_ID","Geography"
"NAME","Geographic Area Name"
"B25070_001E","Estimate!!Total"
"B25070_001M","Margin of Error!!Total"
"B25070_002E","Estimate!!Total!!Less than 10.0 percent"
"B25070_002M","Margin of Error!!Total!!Less than 10.0 percent"
"B25070_003E","Estimate!!Total!!10.0 to 14.9 percent"
"B25070_003M","Margin of Error!!Total!!10.0 to 14.9 percent"
"B25070_004E","Estimate!!Total!!15.0 to 19.9 percent"
"B25070_004M","Margin of Error!!Total!!15.0 to 19.9 percent"
"B25070_005E","Estimate!!Total!!20.0 to 24.9 percent"
"B25070_005M","Margin of Error!!Total!!20.0 to 24.9 percent"
"B25070_006E","Estimate!!Total!!25.0 to 29.9 percent"
"B25070_006M","Margin of Error!!Total!!25.0 to 29.9 percent"
"B25070_007E","Estimate!!Total!!30.0 to 34.9 percent"
"B25070_007M","Margin of Error!!Total!!30.0 to 34.9 percent"
"B25070_008E","Estimate!!Total!!35.0 to 39.9 percent"
"B25070_008M","Margin of Error!!Total!!35.0 to 39.9 percent"
"B25070_009E","Estimate!!Total!!40.0 to 49.9 percent"
"B25070_009M","Margin of Error!!Total!!40.0 to 49.9 percent"
"B25070_010E","Estimate!!Total!!50.0 percent or more"
[...]

PR #280, opened 2024-06-29
PR for #276: Clean `acs_rents`

santiagohermo:
Up to variable `B25070` this task is ready for review @diegogentilepassaro


base/acs_rents/code/build.R
```diff
@@ -0,0 +1,110 @@
...
+
+source('../../../lib/R/save_data.R')
+
+main <- function() {
+  instub  <- '../../../drive/raw_data/acs_rents'
+  outstub <- '../../../drive/base_large/acs_rents'
+  
+  folder_names <- list.files(instub, pattern = "B2*")
+  
+  # These data are aggregated at the block group level
```
diegogentilepassaro, line 12: This comment and the fact that `geo_id` is a key in the save data confused me a bit! so `geo_id` indexes census block groups?

diegogentilepassaro (approved):
All looks good! So the plan is to now use the census block master to aggregate this data to the zcta weighting as usual by number of housing units?

base/acs_rents/code/build.R
diegogentilepassaro: I was expecting data to be at the zcta :/ haha

santiagohermo: Thanks @diegogentilepassaro! Adding a bit more context for the comment:

https://github.com/diegogentilepassaro/min_wage_rent/blob/788d9037221d5f5eabd80210ed43d4f4650d9424/base/acs_rents/code/build.R#L12-L14

The data on gross rent as a % of household income is the one that is at the block level! That's why I exclude it from the list. The other ones are at the zcta.

For now I skipped it to get this started, but if we want to add it we need to work a bit more.

santiagohermo: I improved the comment in https://github.com/diegogentilepassaro/min_wage_rent/pull/280/commits/ba2a3fafc99abaa5f46d9a652d6fda1b1aa0124a @diegogentilepassaro

Issue #281, opened 2024-07-05
Update minimum wage data to 2023


santiagohermo:
In this issue we want to update the minimum wage data until December 2023.

- The raw data is in [this folder](https://drive.google.com/drive/u/0/folders/1oi4guGWvf3Pls5egwIk5cYHuqzxzsnWr).
- For every jurisdiction the last row must be December 2023 (even if it repeats the last minimum wage)

santiagohermo:
Update completed for state-level minimum wage policies.

The last entry for each state is 2024, Dec 31.

Next: update sub-state MW policies.

santiagohermo:
For substate policies:
- Update Arizona, Minnesota, and Washington states
- California on the way

santiagohermo:
Interesting quote from Iowa governor when he blocked local minimum wage policies.

> "I think that uniformity (in wages) is critically important, especially when you have communities that are in more than one county," Branstad said.

Source: https://www.desmoinesregister.com/story/news/politics/2017/03/30/branstad-signs-bills-limiting-workers-compensation-blocking-minimum-wage-hikes/99819450/

@diegogentilepassaro @gabrieleborg

santiagohermo:
Some municipalities opted out of Cook County's minimum wage, should we explore this further?
https://www.arisechicago.org/county

santiagohermo:
Continues in #285

PR #285, opened 2024-07-23
PR for #281: Update minimum wage data to 2023

santiagohermo:
Who wants to review #281 @diegogentilepassaro @gabrieleborg ? I re-run `base/min_wage` after updating the [minimum wage data in the datastore](https://drive.google.com/drive/u/0/folders/1oi4guGWvf3Pls5egwIk5cYHuqzxzsnWr).

Some things to note:
- I added some new places and removed some that were not necessary, see [here](https://github.com/diegogentilepassaro/min_wage_rent/issues/281#issuecomment-2241947130) for details.
- All entries in the raw data have as last date the 31 of December of 2024 or later.
- I realized that most municipalities opted out of the Cook County minimum wage in 2017, see [here](https://github.com/diegogentilepassaro/min_wage_rent/issues/281#issuecomment-2244115151). I think we could only apply the Cook County MW only for the localities that actually use it (if the name of the place in LODES coincides with the name of the municipality), but that is work for a separate issue.


gabrieleborg:
@santiagohermo great! Will review right now

gabrieleborg (approved):
Clean job - thanks for the updates! Regarding Cook County, I have quckly looked at some stats based on the new Census data (2021 stats): 

Screenshot 2024-07-28 at 3 10 04 PM

While Cook county is just a small percentage of the US in terms of population and housing units, It still has 2.6M Low income workers.  So it may be good to open a task to review the MW adoption at the municipality level,

santiagohermo:
Thanks for the quick review @gabrieleborg!

I agree, let's open up a new issue to at least explore the cook county MW issue. To me it sounds like manually adding this info in the construction of the MW panels shouldn't be all that hard, we'll see.

Issue #275, opened 2024-06-14
Get ACS rents data


santiagohermo:
It exists @diegogentilepassaro @gabrieleborg ! Very exciting.

I asked for Census help and I got this response

> Hi Santiago,
> Thank you for reaching out to the U.S. Census Bureau, and I apologize for the delayed response. 
> The best we can offer are the tables below. Lower is the 25th percentile and upper is the 75th percentile. I've also filtered to all ZCTAs in the U.S., hence why it says it's too big to display, but they can download. There will be ZCTAs without estimates, as I'm sure many hit the various filtering rules for display.

> 
 > [B25057: Lower Contract Rent ... - Census Bureau Tables](https://data.census.gov/table?q=B25057:%20Lower%20Contract%20Rent%20Quartile%20%28Dollars%29&g=010XX00US$8600000)
> [B25058: Median Contract ... - Census Bureau Tables](https://data.census.gov/table?q=B25058:%20Median%20Contract%20Rent%20%28Dollars%29&g=010XX00US$8600000)
> [B25059: Upper Contract Rent ... - Census Bureau Tables](https://data.census.gov/table?q=B25059:%20Upper%20Contract%20Rent%20Quartile%20%28Dollars%29&g=010XX00US$8600000)

> 
> 

> If you want to go the extra mile to get a mean, you would have to use table B25060 and B25056. You would divide the aggregate amount found in B25060 by the "Wish cash rent" estimate total. 

> 
> I hope this helps. Let me know if there is anything else I can do to help.
> Thanks,
> Ryan

I'll add the data to our repository in this issue. We need to discuss how to clean it up and add it to the pipeline.

santiagohermo:
Data are available in [`GDrive/raw_data/acs_rents/`](https://drive.google.com/drive/u/0/folders/1R54Ap0Ea1tFwPt3ocnNb7nEGiAKj8bO3)

gabrieleborg:
Hey guys! this is great! @santiagohermo let us know if you want to meet sometime to discuss what we need to do :)

santiagohermo:
Thanks @gabrieleborg! I agree, time to discuss. I'll be in touch :)

santiagohermo:
[...]

Issue #274, opened 2024-06-03
Get IRS SOI data by AGI bracket


santiagohermo:
In this issue we want to expand our IRS data by:

1. First downloading the IRS SOI data by AGI bracket. Source: https://www.irs.gov/statistics/soi-tax-stats-individual-income-tax-statistics-zip-code-data-soi 
2. Create `base/irs_soi/code/build_by_bracket.do` to clean the data, following the procedures we implement in `base/irs_soi/code/build.do`. 
   - The output should be a dataset in `base/irs_soi` called `irs_zip_by_bracket.csv` keyed at the zipcode-year-bracket level. 
   - It should contain the same variables (as long as possible) as [`irs_zip.csv`](https://github.com/diegogentilepassaro/min_wage_rent/blob/1f2abe5369a006c256a2dbbf90ad122a45f71086/base/irs_soi/output/data_file_manifest.log)

santiagohermo:
Plan @gabrieleborg:
- Change `base/irs_soi` to use python 3
- Install https://github.com/gslab-econ/gslab_make through `pip` in a python3 environment
- Update progressively other folders, re-run everything in a new issue with updated templates

santiagohermo:
Continues in #288

PR #286, opened 2024-07-28
PR for #275: acs data

gabrieleborg:
The PR contains all the updates to the scripts downloading ACS Census data: 
1. Added `raw/acs_zcta/code/build.py` script to fetch, combine, rename and store ACS data; 
2. ACS data is stored in `drive/raw_data/acs_rents/acs_zcta_data.csv`; 
3. Removed folder `base/acs_rents/`. Since we've automated the fetching step, we now handle table formatting in the same script (`raw/acs_zcta/code/build.py`); 
4. Related to point 3, all columns that were present in base/acs_rents/  were added maintaining the same naming format that was previously used;

Some comments on the data downloaded: 
5. Table B20026: I have not added it, but if we need there’s also the population in occupied units by time of moving (2021-2022, 2018-2020, etc). This could give us an estimate of movers, and when they moved; 
6. Table B25070: Added both total share and binned distribution for the gross rent as pct of hh income (past 12months). In the name, upper bound is excluded (e.g., `rent_share_hh_income_10_15pct` captures `[10, 14.9]`.) ; 
7. Currently, I haven't added Margin of Errors. Easy to add if we need them. 

General Consderations: 
8. At the moment I left the API key hardcoded in the script. We should consider better management practices for configs and parameters


raw/acs_zcta/code/build.py
```diff
@@ -0,0 +1,324 @@
...
+    # Rent Data
+    'B25026': ['B25026_001E', 'B25026_002E', 'B25026_009E'],  # Population in Occupied Units 
+    'B25057': ['B25057_001E'],  # Lower Contract Rent Quartile (Dollars) 
+    'B25058': ['B25058_001E'],  # Median Contract Rent Quartile (Dollars) 
+    'B25059': ['B25059_001E'],  # Upper Contract Rent Quartile (Dollars) 
+    'B25060': ['B25060_001E'],  # Aggregate Contract Rent (Dollars) 
+    'B25070': ['B25070_001E', 'B25070_002E', 'B25070_003E', 
+               'B25070_004E', 'B25070_005E', 'B25070_006E', 
+               'B25070_007E', 'B25070_008E', 'B25070_009E', 
+               'B25070_010E', 'B25070_011E'],  # Gross Rent as a Percentage of Household Income in the Past 12 Months
```
santiagohermo, line 56: I thought these were aggregated at the census block level. Was I wrong?

raw/acs_zcta/code/build.py
```diff
@@ -0,0 +1,324 @@
...
+        logging.info(f"Final Columns for {year}: {combined_df.columns}")
+
+        return combined_df
+    else:
+        return None
+    
+
+@sleep_and_retry
+@limits(calls=CALLS_PER_SECOND, period=1)
+def fetch_data_for_zctas(year, table, variables, offset=0):
```
santiagohermo, line 115: Fancy...

raw/acs_zcta/code/build.py
```diff
@@ -0,0 +1,324 @@
...
+        os.path.dirname(
+            os.path.dirname(
+                os.path.dirname(
+                    os.path.realpath(__file__)))))
+    
+    data_path  = os.path.join(root_path, "drive", "raw_data", "acs_rents")
+    
+    # Initialize config files
+    logging.basicConfig(filename = os.path.join(output_path, 'build.log'), filemode = 'w', 
+                        format = '%(asctime)s %(message)s', level = logging.INFO, datefmt = '%Y-%m-%d %H:%M:%S')
```
santiagohermo, line 272: I do not see the log file in the data, is it possible that it was not added?

raw/acs_zcta/code/build.py
```diff
@@ -0,0 +1,324 @@
...
+    missing_columns = [col for col in expected_columns if col not in df.columns]
+    if missing_columns:
+        logger.warning(f"The following expected columns are missing from the dataset: {missing_columns}")
+
+    index_cols = ['zcta', 'Year']
+    return df[index_cols + [x for x in df.columns if x not in index_cols]]
+
+
+# test_data_availability()  # Uncomment and Run if needed
+
```
santiagohermo, line 252: Duplicated (the exact same commented code is under the main function.

```suggestion

```

raw/acs_zcta/smsa_fips.txt
santiagohermo, line 1: Why was this file moved? It looks like an error.

santiagohermo (commented):
Thanks @gabrieleborg! This is great :)

I left some questions below.

On your general points:
- 1-4) All makes sense! Can we store the downloaded data in `raw_data/acs_zcta`? Btw, [`raw_data/acs_rents`](https://drive.google.com/drive/u/0/folders/1R54Ap0Ea1tFwPt3ocnNb7nEGiAKj8bO3) seems to have the data I downloaded manually, so we can remove it
- On 5-7. Your decisions make sense to me!
- I commented in-line on 8.

base/acs_rents/output/data_file_manifest.log
```diff
@@ -1,18 +0,0 @@
-
-========================================================================== 
-File: ../../../drive/base_large/acs_rents/zcta_year.csv
```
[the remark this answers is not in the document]
gabrieleborg: dropped it

.gitignore
```diff
@@ -42,3 +42,5 @@ paper_slides/slides/long_slides/min_wage_rent.log
 paper_slides/slides/short_slides/min_wage_rent_talk.log
 paper_slides/slides/short_slides/state-*
 paper_slides/staged_output/output.log
+
+.vscode/
```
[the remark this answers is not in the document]
gabrieleborg: line added

raw/acs_zcta/code/build.py
```diff
@@ -0,0 +1,324 @@
...
+import numpy as np
+from typing import Union
+import glob
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+# Your API key
+API_KEY = 'f0aadefd118cba2a1047e3f4be068bb384908cb2'
```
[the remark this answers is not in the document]
gabrieleborg: Yeah, we can keep it here for the moment :) Ideally yes, you'd keep it in a file that is ignored but let's focus on more important things

raw/acs_zcta/code/build.py
gabrieleborg: Good catch

raw/acs_zcta/smsa_fips.txt
gabrieleborg: Weird I see it in there.

raw/acs_zcta/code/build.py
gabrieleborg: Spent one hour trying to figure out why the log file is not saved. Couldn't find a solution. @santiagohermo @diegogentilepassaro if you guys agree I'll proceed and merge to mainline anyway leaving a note in the script. After that I'll submit a PR for #274

raw/acs_zcta/code/build.py
santiagohermo: That's great @gabrieleborg, thanks for double checking! I guess I was wrong then :)

santiagohermo:
Thanks @gabrieleborg!

- It looks like there is not commit from you, which means [this](https://github.com/diegogentilepassaro/min_wage_rent/pull/286#discussion_r1696917660), [this](https://github.com/diegogentilepassaro/min_wage_rent/pull/286#discussion_r1696928249), and [this](https://github.com/diegogentilepassaro/min_wage_rent/pull/286#discussion_r1696929556) were not changed. Maybe you forgot to push them?
- On [this](https://github.com/diegogentilepassaro/min_wage_rent/pull/286#discussion_r1700173815): it's fine if there is no log file. I would like to double check the data you downloaded before merging though. As I noticed in the [previous comment](https://github.com/diegogentilepassaro/min_wage_rent/pull/286#pullrequestreview-2207613200), the downloaded data are not online. Could you:
  - Add the downloaded data in `GDrive/raw_data/acs_zcta`, and change the code so that this is the output folder?
  - Remove from the GDrive the raw data on `acs_rents` I downloaded previously, which is [here](https://drive.google.com/drive/u/0/folders/1R54Ap0Ea1tFwPt3ocnNb7nEGiAKj8bO3)

raw/acs_zcta/smsa_fips.txt
santiagohermo: Can you bring it back to `raw/ahs`, its original location?

gabrieleborg:
That's weird because I did it and I see only the new folder and new data....
Screenshot 2024-08-02 at 8 45 07 AM

I'll double check what's going on.

raw/acs_zcta/smsa_fips.txt
gabrieleborg: I didn't see any `raw/ahs/` folder. So I created it and moved the `smsa_fips.txt` file in it.

gabrieleborg:
OK found out what was happening @santiagohermo...for some reason there were 2 alternative versions of the GDrive, and I was pointing to the other one. So I was the only one seeing all I was doing 😂. I have removed my symbolic link to the "alternative reality GDrive", and re-added it pointing to the one you also see. I have then re-run the script, created the ACS data - which now you should see. Please review and let me know. AND, for some reason now I there's the log!!! I have also moved the `smsa_fips.txt` to `raw/ahs/`. Let me know about that too. Thanks!

raw/acs_zcta/code/build.py
```diff
@@ -0,0 +1,318 @@
...
+        'B19001_008E': 'Income_35000_39999',
+        'B19001_009E': 'Income_40000_44999',
+        'B19001_010E': 'Income_45000_49999',
+        'B19001_011E': 'Income_50000_59999',
+        'B19001_012E': 'Income_60000_74999',
+        'B19001_013E': 'Income_75000_99999',
+        'B19001_014E': 'Income_100000_124999',
+        'B19001_015E': 'Income_125000_149999',
+        'B19001_016E': 'Income_150000_199999',
+        'B19001_017E': 'Income_200000_or_more',
```
santiagohermo, line 176: This is household income, right? My only comment is that it would be useful to call them "hh_income_.." or something, so that it's clearer. But I don't want to make you run everything again!

raw/acs_zcta/output/build.log
```diff
@@ -0,0 +1,977 @@
...
+2024-08-02 09:07:41 Attempting to fetch data for year 2020, table B00002, geo county:*, offset 0
+2024-08-02 09:07:41 Error fetching data for year 2019, table B00001, geo county:*, offset 0: 400 Client Error:  for url: https://api.census.gov/data/2019/acs/acs5?get=NAME,B00001_001E&for=county:*&key=f0aadefd118cba2a1047e3f4be068bb384908cb2&offset=0
+2024-08-02 09:07:41 Attempting to fetch data for year 2019, table B00001, geo state:*, offset 0
+2024-08-02 09:07:41 Error fetching data for year 2020, table B00002, geo county:*, offset 0: 400 Client Error:  for url: https://api.census.gov/data/2020/acs/acs5?get=NAME,B00002_001E&for=county:*&key=f0aadefd118cba2a1047e3f4be068bb384908cb2&offset=0
+2024-08-02 09:07:41 Attempting to fetch data for year 2020, table B00002, geo state:*, offset 0
+2024-08-02 09:07:41 Error fetching data for year 2019, table B00001, geo state:*, offset 0: 400 Client Error:  for url: https://api.census.gov/data/2019/acs/acs5?get=NAME,B00001_001E&for=state:*&key=f0aadefd118cba2a1047e3f4be068bb384908cb2&offset=0
+2024-08-02 09:07:41 Failed to fetch data for year 2019, table B00001 at all geographic levels
```
santiagohermo, line 468: There were some errors here. Is it something to be concerned about?

raw/acs_zcta/output/build.log
```diff
@@ -0,0 +1,977 @@
...
+2024-08-02 09:08:54 Number of rows in the final dataset: 398748
+2024-08-02 09:08:54 Sample of the final dataset:
+    zcta  Year         NAME  Total_Population state  Total_Housing_Units  \
+0  01832  2013  ZCTA5 01832             22121    25                 7016   
+1  01833  2013  ZCTA5 01833              8295    25                16558   
+2  01834  2013  ZCTA5 01834              6675    25                23738   
+3  01835  2013  ZCTA5 01835             13527    25                 2502
```
santiagohermo, line 923: If you decide to run it again I think we can delete the NAME column.

raw/acs_zcta/output/build.log
```diff
@@ -0,0 +1,977 @@
...
+2024-08-02 09:08:54 Number of rows in the final dataset: 398748
+2024-08-02 09:08:54 Sample of the final dataset:
+    zcta  Year         NAME  Total_Population state  Total_Housing_Units  \
```
santiagohermo, line 919: If you decide to run again I wonder whether we can save a simple data manifest. Like just compute the average and std error of each variable and save it in a log file

santiagohermo (approved):
Thanks @gabrieleborg! Yeah, the old GDrive is still active. I think my @brown account will die this month, and probably the GDrive with it. Yes! Delete old references to the old one please.

Al looks good with the branch now, so I'm approving. I left a couple of comments that I think require a new run, but I don't feel too strongly about it if you think otherwise. Feel free to S&M afterwards!

raw/acs_zcta/code/build.py
gabrieleborg: done

raw/acs_zcta/output/build.log
gabrieleborg: Thanks for pointing that out. I double checked and I found out that (unfortunately for us) the zcta-level data on the unweighted sample count of the population and housing ends at 2018. After that there are only tables at the county or places level. So we have the years 2019-2022 missing for the `Unweighted_Sample_Count_Population` and `Unweighted_Sample_Count_Housing` variables.

raw/acs_zcta/output/build.log
gabrieleborg: done

raw/acs_zcta/output/build.log
gabrieleborg: added a function to create a file manifest (didn't use gslab since I couldn't find file manifest function to use directly with python, as opposed to Stata, R, etc.)

gabrieleborg:
@santiagohermo  I answered all your questions/requests and merged the PR!

PR #288, opened 2024-08-06
PR for #274 - Adding IRS data by brackets. 

gabrieleborg:
Hey @santiagohermo @diegogentilepassaro, 

Here are the updated scripts and files storing IRS tax data. 

1. I have successfully updated the `base/irs_soi/make.py` script to work with gslab_make2. We now just need to import all gslab modules via pip at the top of the script (`from gslab_make import *`). It works mostly simlar to the old version. The main difference is that almost all functions now require a path dictionary storing all the log, input, output paths etc. You can read all details in the [gslab webpage](https://gslab-econ.github.io/gslab_make/index.html)
2. I have then created a `base/irs_soi/code/build_by_bracket.do` based on the `base/irs_soi/code/build.do` original file. In the new file we now have a wide dataset where we have the same variable as in the old data, but expanded for each of the 6 IRS brackets: e.g., in the irs_zip.dta we have `agi_per_hhld` while in the `irs_by_bracket_zip.dta` we have `agi_per_hhld1`-`agi_per_hhld6`, and so on. You can find all stats in the [data file manifest committed](https://github.com/diegogentilepassaro/min_wage_rent/blob/issue_274_irs_soi_data/base/irs_soi/output/data_file_manifest.log).
3. One of the slight differences between the two dofiles is that in the new `build_by_brackets.do` I removed the command `drop if num_hhlds_irs == 0`. While in the original file it made sense to remove zip codes with no taxpayers, we wanted to keep them if, say, they have `num_hhlds_irs1 == 0` because on the other hand the same zip code may have `num_hhlds_irs3 != 0`. In this way when we perform the rehsape wide we correctly have zeros and not missings. Do you think it make sense to keep those zip codes in the main irs data too, so to ensure comparability between the two?


base/irs_soi/make.py
```diff
@@ -0,0 +1,57 @@
...
+    zf.close()
+
+run_stata(program = 'code/build.do', executable = stata, paths=PATHS)
+run_stata(program = 'code/build_by_bracket.do', executable = stata, paths=PATHS)
+
+clear_dir([PATHS[x] for x in ['temp_dir']])
+
+end_makelog(PATHS)
+
+# input('\n Press <Enter> to exit.')
```
santiagohermo, line 57:
```suggestion
# input('\n Press <Enter> to exit.')

```

base/irs_soi/code/build_by_bracket.do
```diff
@@ -0,0 +1,77 @@
...
+        gen wage_per_hhld`i'      = total_wage`i'/num_hhlds_irs`i'*1000
+        gen wage_per_cap`i'       = total_wage`i'/pop_irs`i'*1000
+
+        gen bussines_rev_per_owner`i' = total_bizinc`i'/num_bus_hhlds_irs`i'*1000
+
+        gen share_wage_hhlds`i'      = num_wage_hhlds_irs`i'/num_hhlds_irs`i'
+        gen share_bussiness_hhlds`i' = num_bus_hhlds_irs`i'/num_hhlds_irs`i'
+        gen share_farmer_hhlds`i'    = num_farm_hhlds_irs`i'/num_hhlds_irs`i'
+
+    }
```
santiagohermo, line 73: Can we do something to improve the tag of these variables? I think we want `agi_per_hhld_1_to_25k` instead of `agi_per_hhld1`, etc.

santiagohermo (changes requested):
Awesome work @gabrieleborg! Thanks :)

I think it looks great, I only have one little comment: can we use better tags to name the variables? Example: instead of `agi_per_hhld1`, `agi_per_hhld2`, etc can we have `agi_per_hhld_1_to_25k`, `agi_per_hhld_25k_to_50k`. The list of brackets is [here](https://github.com/diegogentilepassaro/min_wage_rent/issues/274#issuecomment-2241168308). This change should simplify remembering what the brackets are in the future.

On your general comments:
- Awesome work on moving to python 3! We need to implement this more widely. I opened up #289 for this
- `build_by_bracket.do` looks great! Only comment is above. Did you add the new data to the Google Drive?
- According to the data manifests, the same number of zipcodes is available in both datasets. So I think we don't need to worry about those `drop if` statements.

